PR #2, opened 2020-12-14
Adds a base config setter

### LINKED ISSUE #6: Create library configuration
The library should read the address of the workload api socket from an ENV VAR

Unit-tests for should be introduced as part of implementing this issue.

mchurichi:
Closes #6 

Signed-off-by: Maximiliano Churichi <maximiliano.churichi@hpe.com>

### DIFF
--- a/src/pyspiffe/config.py
+++ b/src/pyspiffe/config.py
@@ -1,0 +1,21 @@
+import os
+from typing import Dict
+
+SPIFFE_ENDPOINT_SOCKET = 'SPIFFE_ENDPOINT_SOCKET'
+
+
+class ConfigSetter:
+    def __init__(self):
+        self._apply_default_config()
+        self._apply_environment_variables()
+
+    def get_config(self) -> Dict:
+        return self._config
+
+    def _apply_default_config(self):
+        self._config = {
+            SPIFFE_ENDPOINT_SOCKET: None,
+        }
+
+    def _apply_environment_variables(self):
+        self._config[SPIFFE_ENDPOINT_SOCKET] = os.environ.get(SPIFFE_ENDPOINT_SOCKET)
--- a/src/pyspiffe/config.py
+++ b/src/pyspiffe/config.py
@@ -8,6 +8,26 @@ SPIFFE_ENDPOINT_SOCKET = 'SPIFFE_ENDPOINT_SOCKET'
 
 
 class ConfigSetter:
+    """
+    Loads and validates configuration variables from the environment
+
+    Raises:
+        ValueError: if any variable from the environment has an invalid format
+    """
+
+    FORBIDDEN_SOCKET_COMPONENTS = [
+        ('fragment', None),
+        ('username', None),
+        ('password', None),
+        ('query', None),
+    ]
+
+    UNIX_FORBIDDEN_SOCKET_COMPONENTS = FORBIDDEN_SOCKET_COMPONENTS + [
+        ('netloc', 'authority')
+    ]
+
+    TCP_FORBIDDEN_SOCKET_COMPONENTS = FORBIDDEN_SOCKET_COMPONENTS + [('path', None)]
+
     def __init__(self):
         self._apply_default_config()
         self._apply_environment_variables()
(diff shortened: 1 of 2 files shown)


Andres-GC (commented):
I think we want to add some logic to explicitly raise an exception when the ENV VAR is not set. 
 `SPIFFE_ENDPOINT_SOCKET` is required for the library to work so raising the exception immediately looks like the correct path for me.

test/test_config.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def test_read_value_from_environment_variables():
+    fake_socket = 'unix:///path/to/endpoint.sock'
+    os.environ['SPIFFE_ENDPOINT_SOCKET'] = fake_socket
+
+    setter = ConfigSetter()
+    del os.environ['SPIFFE_ENDPOINT_SOCKET']
+
+    assert setter.get_config()['SPIFFE_ENDPOINT_SOCKET'] == fake_socket
```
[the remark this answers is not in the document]
mchurichi: > I think we want to add some logic to explicitly raise an exception when the ENV VAR is not set.
> `SPIFFE_ENDPOINT_SOCKET` is required for the library to work so raising the exception immediately looks like the correct path for me.

it makes sense!, I'll add that

mchurichi:
> I think we want to add some logic to explicitly raise an exception when the ENV VAR is not set.
> `SPIFFE_ENDPOINT_SOCKET` is required for the library to work so raising the exception immediately looks like the correct path for me.

it makes sense! I'll add that

test/test_config.py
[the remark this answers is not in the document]
mchurichi: I thought it was working but seems I was wrong... I'll double check

src/pyspiffe/config.py
```diff
@@ -0,0 +1,21 @@
...
+    def get_config(self) -> Dict:
+        return self._config
+
+    def _apply_default_config(self):
+        self._config = {
+            SPIFFE_ENDPOINT_SOCKET: None,
+        }
+
+    def _apply_environment_variables(self):
+        self._config[SPIFFE_ENDPOINT_SOCKET] = os.environ.get(SPIFFE_ENDPOINT_SOCKET)
```
glaucimar, line 21: I suggest the following is also checked: 1) environment variable is set;  2) value follow expected format "unix://...." or "tcp://..." and 3) The value set in the environment variable respects maximum permitted length of a path (UNIX_PATH_MAX) in linux.

I wonder if we should also check for the other validations mentioned by SPIFFE specification e.g: if unix scheme, authority component MUST NOT be checked, only scheme and paths are present... and for tcp, host authority MUST be set to an IP address, and make sure scheme, host and port components are set.

If the decision is to follow the path for the extra validation scenario, I would suggest creating a SpiffeEndpointSocket to hold all of these and ConfigSetter could make usage of it.

maxlambrecht: I agree with @glaucimar, we should validate the address following the spiffe spec: https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE_Workload_Endpoint.md#4-locating-the-endpoint
Also it's a good idea moving the validations to another class. In the go-spiffe and java-spiffe there are tests enforcing those constraints.

mchurichi: thanks folks, it makes sense to add these validations... the only thing I'm not sure is if we need to check if the env var is set... the SPIFFE spec says `Clients may be explicitly configured with the socket location, or may utilize the well-known environment variable SPIFFE_ENDPOINT_SOCKET. If not explicitly configured, conforming clients MUST fall back to the environment variable.` what makes me think that the client can be explicitly instantiated with a given socket, and the env var is not always mandatory

test/test_config.py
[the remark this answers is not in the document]
mchurichi: double checked and the coding format seems to be ok, there's already a trailing line break in this file and the pre commit hook is properly set up... marking the comment as resolved, but please reopen if you find anything else

src/pyspiffe/config.py
scotte, line 21: I think this should raise an additional exception with specific details, otherwise the exception isn't very specific on what's wrong and how to resolve - you'll get something like:

```
Traceback (most recent call last):
  File "/home/scott/.pyenv/versions/3.6.12/lib/python3.6/site-packages/pyspiffe/workloadapi/default_workload_api_client.py", line 34, in __init__
    self.spiffe_socket_path=os.environ['SPIFFE_ENDPOINT_SOCKET']
  File "/home/scott/.pyenv/versions/3.6.12/lib/python3.6/os.py", line 669, in __getitem__
    raise KeyError(key) from None
KeyError: 'SPIFFE_ENDPOINT_SOCKET'
```

Which isn't at all clear that this is for an environment variable, or that the variable is a requirement when called this way.

So in my stub implementation (where I implemented a temporary version of this until this code is in), I wrapped that in a RuntimeError (not sure if that's right, but that's what I did), so in addition to the above you get the following with details on what went wrong and how to fix:

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "test.py", line 6, in <module>
    wlapi = DefaultWorkloadApiClient()
  File "/home/scott/.pyenv/versions/3.6.12/lib/python3.6/site-packages/pyspiffe/workloadapi/default_workload_api_client.py", line 36, in __init__
    raise RuntimeError('SPIFFE_ENDPOINT_SOCKET environment variable not specified to DefaultWorkloadApiClient')
```

scotte: Seems reasonable to enforce a path per the SPIFFE specification.

mchurichi: Hey @scotte, this line of code is using the `get` method from the dict type (`os.environ.get`), which doesn't raise an exception when the key doesn't exists, but returns a default value (`None` if not specified by the second parameter). In the other hand, when you try to access the dict by using the a key as an index it does raises an exception, as you stated in the example (`os.environ['SPIFFE_ENDPOINT_SOCKET']`).

Also, as we talked on the last py-spiffe sync meeting, the `SPIFFE_ENDPOINT_SOCKET` is not always mandatory, since a library user can programmatically define the socket by passing it as a parameter when instantiating the API client. So I don't think this is the right place to make such validation.

scotte: I wasn't clear that I understand SPIFFE_ENDPOINT_SOCKET isn't a requirement when programmatically provided - but the point of this method is not to handle that full logic, but simply retrieve the variable if it exists.

So, I think you are right - while raising an exception if the socket isn't provided via any way, that logic probably belongs either in the WorkloadAPI implementation or a new method with the logic for this, but does not belong here. I would argue that as we currently only have one WorkloadAPI implementation anyway, the default one, we handle the runtime logic there and leave the library alone.

Additionally, it's possible that a different WorkloadAPI implementation would want to do something different anyway.

Long story short: It's good as is. 🏁

src/pyspiffe/config.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def _apply_default_config(self):
+        self._config = {
+            SPIFFE_ENDPOINT_SOCKET: None,
+        }
+
+    def _apply_environment_variables(self):
+        endpoint_socket = os.environ.get(SPIFFE_ENDPOINT_SOCKET)
+
+        if endpoint_socket is not None:
```
Andres-GC, line 26: nit: `endpoint_socket` is of type `str` and strings are falsy (None and empty strings evaluates to False) in python so this can be reduced to:
```
if endpoint_socket:
    <code>
```

src/pyspiffe/config.py
```diff
@@ -0,0 +1,90 @@
...
+
+class ConfigSetter:
+    def __init__(self):
+        self._apply_default_config()
+        self._apply_environment_variables()
+
+    def get_config(self) -> Dict:
+        return self._config
+
+    def _apply_default_config(self):
```
Andres-GC, line 18: nit: let's be explicit on the return type `-> None` in all pleases please. That helps to be consistent with other modules.

src/pyspiffe/config.py
```diff
@@ -0,0 +1,90 @@
...
+
+        if parsed_socket.scheme == '':
+            raise ValueError('SPIFFE endpoint socket: scheme must be set')
+
+        if parsed_socket.scheme == 'unix':
+            SpiffeEndpointSocket.__validate_unix_socket(parsed_socket)
+        elif parsed_socket.scheme == 'tcp':
+            SpiffeEndpointSocket.__validate_tcp_socket(parsed_socket)
+        else:
+            raise ValueError('SPIFFE endpoint socket: unsupported scheme')
```
Andres-GC, line 44: All exception messages on the [spiffe_id module](https://github.com/HewlettPackard/py-spiffe/blob/master/src/pyspiffe/spiffe_id/spiffe_id.py) ends with a final period. I'll think we should be consistent on that.
For now I'll suggest following that format and if required, we can discuss this topic again.

src/pyspiffe/config.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def _apply_environment_variables(self):
+        endpoint_socket = os.environ.get(SPIFFE_ENDPOINT_SOCKET)
+
+        if endpoint_socket is not None:
+            SpiffeEndpointSocket.validate(endpoint_socket)
+            self._config[SPIFFE_ENDPOINT_SOCKET] = endpoint_socket
+
+
+class SpiffeEndpointSocket:
```
Andres-GC, line 31: Would you consider adding a docstring to each new class please?

src/pyspiffe/config.py
mchurichi: thanks!, good tip 👌

scotte (approved):

scotte (approved):

Andres-GC (approved):
:medal_sports:  Thanks Max

PR #3, opened 2020-12-17
Implement TrustDomain. Add pytest dependency

maxlambrecht:
Implementing TrustDomain strong type.


src/pyspiffe/__init__.py
```diff
@@ -0,0 +1 @@
+SPIFFE_SCHEME = 'spiffe'
```
mchurichi, line 1: s/SPIFFE_SCHEME/SPIFFE_SCHEMA

src/pyspiffe/spiffe_id/trust_domain.py
```diff
@@ -1,11 +1,59 @@
...
+        uri = urlparse(name)
+        validate_uri(uri)
+        self.__name = uri.hostname.lower().strip()
+
+    def get_name(self):
+        return self.__name
+
+
+def normalize(name: str):
+    if not name.__contains__(SCHEMA_SUFFIX):
```
mchurichi: what if we check that the name starts with the right schema?
```python
if not name.startswith(SPIFFE_SCHEME + SCHEME_SUFFIX):
```
We could also validate it's a valid url by parsing it, but that's probaby overkill.
Also, `__contains__`, as most of the magic methods, is not intended to be used directory. It's recommended to use `in` instead: `if not XXX in name`:

maxlambrecht: Good catch! I replaced the magic method call by the `in` operator. The logic behind this is to add the default schema in case is doesn't have any and then parse the URL and performs the validations on the URL.

src/pyspiffe/__init__.py
maxlambrecht: why `SCHEMA` ?  The URI spec uses `SCHEME`: https://tools.ietf.org/html/rfc3986#page-17

mchurichi: if SCHEME is the right word we should then standarize on it, since we also have [`SCHEMA_SUFFIX`](https://github.com/HewlettPackard/py-spiffe/pull/3/files#diff-b3c9101ea47eaa2dcd58b5603cd29f76d090b1f4c5413272a97bffad4ffcdc07R6)

maxlambrecht: yes, you are right, will make it consisteng

src/pyspiffe/spiffe_id/trust_domain.py
mchurichi: sgtm

src/pyspiffe/spiffe_id/trust_domain.py
```diff
@@ -1,11 +1,59 @@
...
+            name = SPIFFE_SCHEME + SCHEME_SUFFIX + name
+        return name
 
-    def __init__(self, trust_domain: str):
-        self.name = trust_domain
+    @staticmethod
+    def __validate_uri(uri):
+        if not uri.scheme == SPIFFE_SCHEME:
+            raise ValueError(
+                'Trust domain: invalid scheme: expected {}'.format(SPIFFE_SCHEME)
```
mchurichi, line 54: I'll start to be super nit picky so from now on you can ignore me 😝 
what if we try to be consistent and extract this error to a constant, same as you are already doing with `EMPTY_DOMAIN_ERROR`, and the same with `Trust domain: port is not allowed` a few lines below

test/spiffe_id/test_trust_domain.py
```diff
@@ -0,0 +1,62 @@
...
+
+@pytest.mark.parametrize(
+    'test_input,expected',
+    [
+        ('domain.test', 'domain.test'),
+        (' doMain.tesT  ', 'domain.test'),
+        ('spiffe://domAin.Test', 'domain.test'),
+        ('spiffe://domain.test/path/element', 'domain.test'),
+        ('spiffe://domain.test/spiffe://domain.test/path/element', 'domain.test'),
+        ('spiffe://domain.test/spiffe://domain.test:80/path/element', 'domain.test'),
```
mchurichi, line 14: even if we can extract the domain name, is this actually a valid trust domain according to the SPIFFE spec?

src/pyspiffe/spiffe_id/trust_domain.py
maxlambrecht: I didn't extract it because my logic was "extract constant when it is used in multiple places"

test/spiffe_id/test_trust_domain.py
maxlambrecht: Yes,  the domain name is taken from the authority, we don't care about the rest, only that the authority in this case cannot have a port. The rest of the URL is validated in case you want to create a SpiffeID from it.

src/pyspiffe/__init__.py
Andres-GC, line 1: This is at top level. I might be missing something but what about setting this constant on the spiffe_id package?

src/pyspiffe/spiffe_id/trust_domain.py
```diff
@@ -1,11 +1,59 @@
...
+    Raises:
+        ValueError: is the name of the trust domain is empty, has a port, or contains an invalid scheme.
+
     """
 
-    name: str
+    def __init__(self, name: str):
+        self.__set_name(name)
+
+    def __str__(self):
```
Andres-GC, line 24: Missing typing for method return type"
```
def __str__(self) -> str:
```
I know is obvious here, I just use this one because is the first place where is missing,  We should try to set typing for all the code.

src/pyspiffe/spiffe_id/trust_domain.py
```diff
@@ -1,11 +1,61 @@
...
 class TrustDomain(object):
     """
     Represents the name of a SPIFFE trust domain (e.g. 'domain.test').
 
-    :param trust_domain: The name of the Trust Domain
+    Args:
+        name(str): The name of the Trust Domain
+
+    Raises:
+        ValueError: is the name of the trust domain is empty, has a port, or contains an invalid scheme.
```
mchurichi, line 19:
```suggestion
        ValueError: if the name of the trust domain is empty, has a port, or contains an invalid scheme.
```

Andres-GC (approved):
LGTM
Thanks Max

mchurichi (dismissed):
LGTM

Andres-GC (approved):
:100:  Thanks Max!

PR #10, opened 2020-12-17
Add initial workloadapi stub and default values

scotte:
Initial pass at creating an abstract class WorkloadApiClient and a stub implementation in DefaultWorkloadApiClient. This was basically just a copy/paste from java-spiffe, as a result there's a number of questionable bits that we probably don't need.

With an implementation of fetchJwtSvid in DefaultWorkloadApiClient, the following test program works successfully:
```
#!/usr/bin/env python

from pyspiffe.workloadapi.default_workload_api_client import DefaultWorkloadApiClient

wlapi = DefaultWorkloadApiClient.new_client()
wlapi.fetch_jwt_svid(audiences=["foo"])
```

UPDATE: Fixed variable and method names to python standards


glaucimar (commented):
Hi Scott, thanks for the PR. Please consider my comment in the default_workload_api_client.py

src/pyspiffe/internal/defaults.py
```diff
@@ -0,0 +1,4 @@
+class Defaults(object):
+
+    spiffe_socket_path = "/tmp/agent.sock"
```
Andres-GC, line 3: There is an [open PR](https://github.com/HewlettPackard/py-spiffe/pull/2) from @mchurichi that generates a config object for the library. One thing we mentioned there is the idea of throwing an exception at config phase in case the ENV VAR that holds the SPIFFE endpoint is not set. Following that I think we should not use a default value for the `spiffe_socket_path`.

Do you think that is correct?

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -0,0 +1,138 @@
...
+
+        Parameters:
+
+        watcher (?? type): Watcher callback object.
+        """
+
+        pass
+
+
+    def fetch_jwt_svid(self, audiences, subject=None):
```
Andres-GC, line 80: Let's include type hints  to follow [PEP 484](https://www.python.org/dev/peps/pep-0484/)
For example:
```
def greeting(name: str) -> str:
    return 'Hello ' + name
```

src/pyspiffe/internal/defaults.py
scotte: I think we should follow along what is done with the other libraries to be consistent - it looks like go-spiffe expects either `SPIFFE_ENDPOINT_SOCKET ` or an explicitly coded path with no default as well? If I'm reading that correctly, then I agree and I'll remove the default.

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -0,0 +1,138 @@
...
+            self.spiffe_socket_path=spiffe_socket_path
+
+        # TODO defaults for these, or delete them entirely
+        self.exponential_backoff_policy=exponential_backoff_policy
+        self.executor_service=executor_service
+
+
+    # TODO: Is this pythonic? Any benefit of this being a factory object vs
+    # just using the constructor?
+    @classmethod
```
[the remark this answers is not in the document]
scotte: I guess I need to leave this to someone with more recent Python experience. Does Python not allow overriding contructors? Seems like I've done that before, but maybe not.

Since this is pretty fundamental to the design of the API I do think this is something we should decide now. I.e. if we release the first version of the library using a constructor, but then later switch to a factor pattern that would change the API which would be a bad idea.

src/pyspiffe/workloadapi/default_workload_api_client.py
scotte: Thanks, I'll look into this.

src/pyspiffe/workloadapi/default_workload_api_client.py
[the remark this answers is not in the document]
glaucimar: No, it does not allow allow for overriding constructors .. this is when we would want another method to instantiate an object of the class.... Agree with you that changing the API would be a bad idea but I don't see how changing to a factor pattern here would change the API. As far as I am concerned this workload_api_client should not be visible for the clients of py-spiffe. And I am not a fan of doing thing for a future we don't know if will ever come. With that said, I don't think this is a fundamental problem or issue. I have my preference but I could go with either way. So if you feel strongly about keeping it, I am fine.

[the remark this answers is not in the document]
maxlambrecht: The Workload API client _is_ part of the API in the other two libraries, because although the library users will generally prefer to use the X509Source they may sometimes want to make a one-shot call to fetch an SVID, and that functionality is provided by the workload API client.

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -0,0 +1,138 @@
...
+    def new_client(cls, spiffe_socket_path=None, exponential_backoff_policy=None,
```
maxlambrecht, line 26: As we talked offline, I would leave out the retry-functionality-related stuff and implement it as a separate feature.

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -0,0 +1,138 @@
...
+        Fetches a SPIFFE JWT-SVID on one-shot blocking call.
+
+        Parameters:
+
+        audiences (set of String): Set of audiences for the JWT.
+        subject (String) [Optional]: SPIFFE ID Subject for the JWT.
+
+        Returns:
+
+        JwtSvid: Instance of JwtSvid object
```
maxlambrecht, line 91: nit: I'm not sure about the _best_ name in python for `JwtSvid`, as it is a _JWT-SVID_

src/pyspiffe/workloadapi/default_workload_api_client.py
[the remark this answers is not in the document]
glaucimar: Hmmmm very good point @maxlambrecht, I was not considering this usage. Thanks!

scotte:
Added initial tests

src/pyspiffe/workloadapi/default_workload_api_client.py
scotte: Me neither! Open to suggestions...

src/pyspiffe/workloadapi/default_workload_api_client.py
scotte: Should be resolved, but please verify as this is something I haven't used in Python before.

src/pyspiffe/internal/defaults.py
scotte: Resolved. The environment variable name is currently hardcoded, maybe this should move to a string table at some point. For now, no config object is required and Defaults was removed.

Andres-GC (commented):
Thanks Scott. It looks ok. I've added a couple of minor comments.

src/pyspiffe/workloadapi/default_workload_api_client.py
Andres-GC: I've just added a comment indicating that we should also include type information for method parameters. 
Thanks for the patience Scott.

src/pyspiffe/workloadapi/default_workload_api_client.py
mchurichi: I think that according to PEP8 the idiomatic way is to have acronyms in all-caps, but `JWTSVID` isn't probably easier to read. I'd say let's use `JwtSvid`.

Reference: https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles
`Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.`

src/pyspiffe/workloadapi/default_workload_api_client.py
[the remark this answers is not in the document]
scotte: For now I've removed the factory method and left only the constructor. I continue to have no opinion over which is "the right way". It is easier to add one later than take it away, so that's why it's removed for now.

src/pyspiffe/workloadapi/default_workload_api_client.py
scotte: Whoops, I missed the hint on parameters,  I only did the return type. I'll fix this, thanks!

test/workloadapi/test_default_workload_spi_client.py
```diff
@@ -0,0 +1,81 @@
+import os
```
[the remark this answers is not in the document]
scotte: Doh! Thanks, fixed. Glad you have good eyes. :-)

test/workloadapi/test_default_workload_spi_client.py
```diff
@@ -0,0 +1,81 @@
...
+        DefaultWorkloadApiClient()
+
+# With SPIFFE_ENDPOINT_SOCKET, and no path passed, succeeds
+
+
+def test_instantiate_default_with_var():
+    os.environ['SPIFFE_ENDPOINT_SOCKET'] = '/tmp/agent.sock'
+    wlapi = DefaultWorkloadApiClient()
+    del os.environ['SPIFFE_ENDPOINT_SOCKET']
+    assert wlapi.spiffe_socket_path == "/tmp/agent.sock"
```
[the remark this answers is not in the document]
scotte: OK, I'll make them consistent.

scotte:
I'm going to abandon this Pull Request and make a new one since things have changed so much since the earlier reviews.

Issue #4, opened 2020-12-17
Implement spiffe_id module


Andres-GC:
The `spiffe_id` module provides the type for the SPIFFE ID and method to parse and validate an SPIFFE ID. 
Well formed SPIFFE IDs should follow the [SPIFFE ID specification.](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md)

In addition, the trust domain portion of the SPIFFE ID should be normalized to lower-case to facilitate SPIFFE ID comparison.

Unit-tests for all functionality in the `spiffe_id` module should be introduced as part of implementing this issue.

PR #22, opened 2021-01-11
Implement `SpiffeId` type.

maxlambrecht:
Implement the strong type `SpiffeId` following the SPIFFE standard.

Add some missing methods and validations to `TrustDomain` type.

Addresses #4

### DIFF
--- a/src/pyspiffe/spiffe_id/spiffe_id.py
+++ b/src/pyspiffe/spiffe_id/spiffe_id.py
@@ -1,17 +1,165 @@
-from src.pyspiffe.spiffe_id.trust_domain import TrustDomain
+from typing import Any
+
+from pyspiffe.spiffe_id import SPIFFE_SCHEME
+from pyspiffe.spiffe_id.trust_domain import TrustDomain
+
+from rfc3987 import parse
+
+SPIFFE_ID_MAXIMUM_LENGTH = 2048
 
 
 class SpiffeId(object):
     """
-    Represents a SPIFFE ID as defined in the SPIFFE standard.
-    see <a href="https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md">https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md</a>
-
-    :param trust_domain: The trust domain corresponds to the trust root of a system.
-    :param path: The path component
+    Represents a SPIFFE ID
+    as defined in the `SPIFFE standard <https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md>`
     """
 
-    schema: str = 'spiffe'
+    @classmethod
+    def parse(cls, spiffe_id: str):
+        """Parses a SPIFFE ID from a string into a SpiffeId type instance.
 
-    def __init__(self, trust_domain: TrustDomain, path: str):
-        self.trust_domain = trust_domain
-        self.path = path
+        Args:
+            spiffe_id: a string representing the SPIFFE ID
+
+        Returns:
+            an instance of a compliant SPIFFE ID (SpiffeId type)
+
+        Raises:
+            ValueError: if the string spiffe_id doesn't comply the the SPIFFE standard.
+
+        Examples:
+            >>> spiffe_id = SpiffeId.parse('spiffe://domain.test/path/element')
+            >>> print(spiffe_id.trust_domain())
+            domain.test
+            >>> print(spiffe_id.path())
+            /path/element
+        """
+
+        if spiffe_id == '' or spiffe_id is None:
+            raise ValueError('SPIFFE ID cannot be empty.')
+
+        uri = cls.parse_and_validate_uri(spiffe_id)
+
+        result = SpiffeId()
+        result.__set_path(uri.get('path'))
+        result.__set_trust_domain(TrustDomain(uri.get('authority')))
+        return result
+
+    @classmethod
+    def of(cls, trust_domain: TrustDomain, path_segments=None):
+        """Creates SpiffeId type instance from a Trust Domain and zero or more paths.
+
+        Args:
+            trust_domain(TrustDomain): The trust domain corresponds to the trust root of a system.
+            path_segments (optional): can be a single string or a list of path segments
+
+        Returns:
+            an instance of a compliant SPIFFE ID (SpiffeId type)
+
+        Raises:
+            ValueError: if the trust_domain is None or is not instance of the class TrustDomain.
+
+        Examples:
+            >>> spiffe_id_1 = SpiffeId.of(TrustDomain('example.org'), 'path')
+            >>> print(spiffe_id_1)
+            spiffe://example.org/path
+
+            an array of paths:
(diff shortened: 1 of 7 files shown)


scotte:
The SPIFFE spec says that all URIs need to be ASCII. Is this something the other libraries enforce as well? Maybe the  rfc3987 library already handles this, and if not I'm not sure we should just so we are at parity with go-spiffe and java-spiffe. Thoughts?

scotte (approved):

maxlambrecht:
> The SPIFFE spec says that all URIs need to be ASCII. Is this something the other libraries enforce as well? Maybe the rfc3987 library already handles this, and if not I'm not sure we should just so we are at parity with go-spiffe and java-spiffe. Thoughts?

Thanks @scotte for asking! 

From the [SPIFFE spec](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md):

```
 A SPIFFE Identity (or SPIFFE ID) is defined as an RFC 3986 compliant URI comprising a “trust domain name” and an associated path.  
...
RFC 3986 permits only ASCII characters.
```

test/spiffe_id/test_spiffe_id.py
```diff
@@ -0,0 +1,184 @@
...
+        SpiffeId.of('', 'path')
+
+    assert (
+        str(exception.value)
+        == 'SPIFFE ID: trust_domain argument must be a TrustDomain instance.'
+    )
+
+
+def test_exceeds_maximum_length():
+    path = "a" * 2028
```
Andres-GC, line 159: nit: there are some places where double quotes are used instead of single quotes. Let's use only single quotes to be consistent.

Andres-GC (commented):
Thanks Max. Looks good, I just added a couple of minor comments.

src/pyspiffe/spiffe_id/spiffe_id.py
```diff
@@ -1,17 +1,168 @@
...
+        if trust_domain is None:
+            raise ValueError('SPIFFE ID: trust domain cannot be empty.')
+
+        result = SpiffeId()
+        if isinstance(trust_domain, TrustDomain):
+            result.__set_trust_domain(trust_domain)
+        else:
+            raise ValueError(
+                'SPIFFE ID: trust_domain argument must be a TrustDomain instance.'
+            )
```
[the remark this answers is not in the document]
maxlambrecht: I'll go for the second option.

src/pyspiffe/spiffe_id/trust_domain.py
```diff
@@ -26,17 +34,27 @@ def __str__(self) -> str:
...
         return False
 
     def name(self) -> str:
         return self.__name
 
+    def as_str_id(self) -> str:
+        return '{}://{}'.format(SPIFFE_SCHEME, self.__name)
+
     def __set_name(self, name: str):
         if name == '' or name is None:
```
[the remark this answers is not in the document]
maxlambrecht: TIL, thanks!

maxlambrecht:
Thanks @Andres-GC for your useful comments! I've addressed them all.

src/pyspiffe/spiffe_id/spiffe_id.py
```diff
@@ -1,17 +1,168 @@
...
+            return '{}://{}'.format(SPIFFE_SCHEME, self.__trust_domain.name())
+
+    # path_segments can be an array of path segments or a single string representing a path
+    def __set_path(self, path: Any):
+        if isinstance(path, list):
+            self.__path = ''
+            for s in path:
+                self.__path += self.normalize_path(s)
+        else:
+            self.__path = self.normalize_path(path)
```
Andres-GC, line 112: One thing I forgot to mention in my previous comment here. `else` statements in both methods are unnecessary. Let's remove them to make it clearer. Also the return type is missing in the second method. Let's add it please.
```
    def __set_path(self, path: Any) -> None:
        if isinstance(path, list):
            self.__path = ''
            for s in path:
                self.__path += self.normalize_path(s)
            return
         self.__path = self.normalize_path(path)
         return
```

src/pyspiffe/spiffe_id/spiffe_id.py
```diff
@@ -1,17 +1,168 @@
...
+
+    def __set_trust_domain(self, trust_domain: TrustDomain):
```
Andres-GC, line 114: Let's include the return type here too please : `def __set_trust_domain(self, trust_domain: TrustDomain) -> None`

src/pyspiffe/spiffe_id/spiffe_id.py
maxlambrecht: The last `return` in that code suggested is not necessary.

Andres-GC: Yes, you are right.
Thanks Max!

Andres-GC (approved):
Thanks Max :1st_place_medal:

src/pyspiffe/spiffe_id/spiffe_id.py
```diff
@@ -1,17 +1,167 @@
...
+    def path(self) -> str:
+        return self.__path
+
+    def trust_domain(self) -> TrustDomain:
+        return self.__trust_domain
+
+    def is_member_of(self, trust_domain: TrustDomain) -> bool:
+        return self.__trust_domain == trust_domain
+
+    @classmethod
```
glaucimar, line 125: Max, I am wondering why is this a classmethod instead of staticmethod if it does not use any class method nor attribute. And parse method only calls this method from the class too which makes me wonder if both shouldn't be static methods.

src/pyspiffe/spiffe_id/spiffe_id.py
```diff
@@ -1,17 +1,167 @@
...
+        if fragment:
+            raise ValueError('SPIFFE ID: fragment is not allowed.')
+
+        authority = uri.get('authority')
+        if not authority:
+            raise ValueError('SPIFFE ID: trust domain cannot be empty.')
+
+        # has user@info:
+        if '@' in authority:
+            raise ValueError('SPIFFE ID: user info is not allowed.')
```
glaucimar, line 154: Should info be spelled complete instead: information?

glaucimar (commented):
Max, This looks good, I just have a couple of questions for you.

src/pyspiffe/spiffe_id/spiffe_id.py
maxlambrecht: good catch!

src/pyspiffe/spiffe_id/spiffe_id.py
maxlambrecht: amended to `userinfo`, that's the component name of the URI in the RFC 3986.

glaucimar: Thanks Max, much better!

glaucimar (approved):
LGTM

Andres-GC (approved):
Thanks Max.

PR #25, opened 2021-01-12
Update README.md

glaucimar:
Adding "clone your own repo" to the contribution section and making a reference to this section on the "Development setup"  to make it  more clear the steps to be followed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,13 @@ The _RFC - Py-Spiffe features and use cases_ document is public at: https://docs
 * Gomez Coronel Andres (HPE)
 * Lambrecht Max (HPE)
 
+## Contributing
+1. Fork it (https://github.com/HewlettPackard/py-spiffe/fork)
+2. Clone your forked repository (git clone git@github.com:<your_github_account>/py-spiffe.git)
+3. Create your feature branch (git checkout -b feature/fooBar)
+4. Commit your changes (git commit -m 'Add some fooBar' --signoff)
+5. Push to the branch (git push origin feature/fooBar)
+6. Create a new Pull Request
 
 ## Development setup
 
@@ -35,7 +42,7 @@ pyenv install 3.9.x
 
 Clone the repository
 ```
-git clone git@github.com:HewlettPackard/py-spiffe.git
+Follow though steps in Contributing session.
 ```
 
 cd to the py-spiffe directory
@@ -78,13 +85,6 @@ To work around this issue, update the version of `pipenv` to `2020.11.4` or newe
 ```
 pip install --upgrade --user pipenv
 ```
-
-## Contributing
-1. Fork it (https://github.com/HewlettPackard/py-spiffe/fork)
-2. Create your feature branch (git checkout -b feature/fooBar)
-3. Commit your changes (git commit -m 'Add some fooBar' --signoff)
-4. Push to the branch (git push origin feature/fooBar)
-5. Create a new Pull Request
 
 ### Code style guide
 The project follows the Google-style (https://google.github.io/styleguide/pyguide.html)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,7 +42,7 @@ pyenv install 3.9.x
 
 Clone the repository
 ```
-git clone git@github.com:HewlettPackard/py-spiffe.git
+Follow though steps in Contributing session.
```
maxlambrecht, line 45: `Follow through the steps in [Contributing](#Contributing) section`

README.md
```diff
@@ -35,7 +42,7 @@ pyenv install 3.9.x
 
 Clone the repository
 ```
-git clone git@github.com:HewlettPackard/py-spiffe.git
+Follow through the steps in [Contributing](#Contributing) session.
```
mchurichi, line 45: I think this should be `section` instead of `session`

glaucimar: Done, thanks :)

README.md
glaucimar: Done, thanks Max!

maxlambrecht (approved):
💯

mchurichi (dismissed):
LGTM

README.md
```diff
@@ -35,7 +42,7 @@ pyenv install 3.9.x
 
 Clone the repository
 ```
-git clone git@github.com:HewlettPackard/py-spiffe.git
+Follow through the steps in [Contributing](#Contributing) section.
```
Andres-GC, line 45: One detail, the text is inside a code block so the link style doent's work. It looks like this:

image

On a different note, referring to the whole contributing section looks to me like "exaggerated". there are other steps there that are not required for the purpose of just _get the repository cloned_ as the header suggest.

What about something like (just text, no code block style):

Clone the repository
Follow steps 1 and 2 from the [Contributing](#Contributing) section.

Andres-GC (commented):
Thanks Glaucimar. I've added a minor comment for a cosmetic change. As part of that change, would you mind adding Scott to the list of contributors?
`Scott Emmons (HPE)`

README.md
scotte: Personally, I don't see the need to spell out the steps too explicitly - it's no different than any other github repository.

Andres-GC: While I do agree with Scott, I need to mention that the PR was originally created to add step two:
`2- Clone your forked repository (git clone git@github.com:<your_github_account>/py-spiffe.git)`

That is also a common path when working with Github. So if we add that for less experienced persons I think adding the correct steps here will make thing less confusing for them. That was my point here.

Said that, I'm Ok with the text as it is too.

README.md
```diff
@@ -8,11 +8,21 @@ Initial work in progress to create a python library for SPIFFE.
 The _RFC - Py-Spiffe features and use cases_ document is public at: https://docs.google.com/document/d/1IByE9Ge1MTyWD7rL9o99-1c5Fn83p1ra3eApARY9X-I/edit#heading=h.70d940e8vk81
 
 ## Initial contributors
```
mchurichi, line 10: super nit picky, but wondering if we should standarize on "FirstName LastName", my OCD isn't letting me see this 😂 
@Andres-GC @maxlambrecht would you mind?

Andres-GC: +1 to list all names in the same format. 
I don't have any preference between Name-Lastname or Lastname-Name

README.md
glaucimar: Done!

README.md
glaucimar: Thanks folks! Done!

Andres-GC (approved):
:ice_cream:

mchurichi (approved):

PR #26, opened 2021-01-13
Update Travis config to run the test

Andres-GC:
Updates Travis-CI config at `.travis.yml` to run the test for the library on different versions of Python.

Changes:
- Set dist to `focal`. **Question:** Should we consider using an older one?
- Specifies the correct `tox` environment for each Python version
- Runs `tox` directly without calling `make test`

`make test` reminds for running the test locally at development.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,20 @@
 language: python
-python:
-    - "3.6"
-    - "3.7"
-    - "3.8"
-    - "3.9"
+dist: focal
+
+jobs:
+  include:
+    - python: 3.6
+      env: TOXENV=py36
+    - python: 3.7
+      env: TOXENV=py37
+    - python: 3.8
+      env: TOXENV=py38
+    - python: 3.9
+      env: TOXENV=py39
 
 # command to install dependencies
-install: "make dev"
+install: "pip install tox"
 
 # command to run tests
 script:
-    - make test
+    - tox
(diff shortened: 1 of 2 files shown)


scotte:
>  Set dist to focal. Question: Should we consider using an older one?

I think we should always use the latest LTS, so focal is good.

scotte (approved):

.travis.yml
```diff
@@ -1,13 +1,20 @@
...
+jobs:
+  include:
+    - python: 3.6
+      env: TOXENV=py36
+    - python: 3.7
+      env: TOXENV=py37
+    - python: 3.8
+      env: TOXENV=py38
+    - python: 3.9
+      env: TOXENV=py39
```
mchurichi, line 13: does this uses the latest minor version for 3.9?

scotte: That's a good point/question!

scotte (approved):
OK, that was weird.

mchurichi (approved):
:shipit:

.travis.yml
Andres-GC: For the specific case of  `3.9` is using the latest version which is now `3.9.1`.
For the other python versions:
```
3.6.9 (3.6.12 is the latest now)
3.7.7 (3.7.9 is the latest now)
3.8.3 (3.8.7 is the latest now)
```
As Travis updates the Python build images, aliases for each version will point to different exact versions or patch levels.

PR #27, opened 2021-01-19
Add lint Makefile target

scotte:
"make lint" runs flake8 on all sources in the "src" and "test" directories.

Signed-off-by: Scott Emmons <scotte@hpe.com>


scotte:
Do we want linting to be a Makefile dependency for `make test`?

Makefile
```diff
@@ -54,6 +64,25 @@ pb_generate:
...
+
+flake8:
+	@echo "Running flake8."
+	$(PIPENV_CMD) run flake8 --statistics --doctests --ignore E203,E266,E501,W503 src test
+
+mypy:
+	@echo "Running mypy."
+	$(PIPENV_CMD) run mypy -m src
+
+
```
Andres-GC, line 85: :100: for adding these tasks as _internal_

For those not familiarized with the `flake8` options and rules, here are some links.
Rules: https://www.flake8rules.com/
Options: 
```
--statistics          Count errors and warnings.
--doctests            check syntax of the doctests
```
[Source](https://flake8.pycqa.org/en/latest/manpage.html#options)

Makefile
```diff
@@ -32,11 +32,21 @@ dev: env
...
 
 
+## Lint source files.
+lint: black flake8 mypy
+
+
+## Reformat source files with black.
+format:
+	@echo "Running black to format source files."
+	$(PIPENV_CMD) run black src test
```
Andres-GC, line 47: I don't have an strong opinion here. I would like to know what other think about this.

Would it make it clearer to the reader if we include the config file as a command parameter here? Like:
`$(PIPENV_CMD) run black --config pyproject.toml src test`

From [black docs](https://black.readthedocs.io/en/stable/pyproject_toml.html#pyproject-toml) it is clear that when the command is run, it will look for a `pyproject-toml` file on common base directory but not sure it this is something that everybody will think about by just looking at the command.

Andres-GC (approved):
Thanks Scott! :tada:

Makefile
scotte: That's a great idea to explicitly call out the config file. I think the flake8 config can be done in there as well, I'll check.

scotte: OK, done in the latest commit - explicit config file for black. For flake8, it *can* use pyproject.conf, however the flake8 config is not compatible with black so adding flake8 config causes black to fail reading the config! So, instead I added the flake8 configuration to tox.ini - which seems a little strange, but is apparently OK (documented at flake8).

Alternatively, we can move flake8 config out to it's own file, but it being in tox.ini didn't seem too bad. Thoughts?

Andres-GC: I've seen the flake8 config in the tox.ini file in different projects. I prefer that instead of creating a new config file.

scotte: OK, great, thanks!

scotte:
Convert to draft so it can't be merged before I can squash the commits.

Andres-GC (dismissed):

scotte:
Squashed and ready for final review.

Andres-GC (approved):
Thanks Scott :1st_place_medal:

mchurichi (approved):
:shipit:

PR #28, opened 2021-01-20
Feature/jwtsvid

glaucimar:
Partially address  #16 

Changes:
- added PyJWT [crypto] dependency
- added tests for JWTSvid
- implemented parse_insecure
- added specific exceptions 
- added docstrings
- created an additional class JWTSvidValidator

Remaining work to fully address the issue:
- parse_validate
- jwt bundle
- tests for these


src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,156 @@
...
+    
+    Raises:
+        JwtSvidError:   when the token expired or the expiration claim is missing, or 
+                        when the 'aud' has an audience that is not in the audience provided as parameter
+        ValueError:     when the token is blank or cannot be parsed
+    """
+
+    @classmethod
+    def parse_insecure(cls, token: str, audience: []) -> 'JwtSvid':
+        if token is None or token == '':
```
maxlambrecht, line 49: can be simplified as `if not token`:

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,156 @@
...
+            raise JwtSvidError(EMPTY_TOKEN_ERROR)
```
maxlambrecht, line 50: here it's probably best to raise a ValueError, as specified in the docstring

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,156 @@
...
+        self.audience = audience
+        self.expiry = expiry
+        self.claims = claims
+        self.token = token
 
-    :param spiffe_id: SPIFFE ID of the X509-SVID.
-    :param private_key: Audience is the intended recipients of JWT-SVID as present in the 'aud' claim.
-    :param chain: Expiration time of JWT-SVID as present in 'exp' claim.
     """
+    Parses and validates JWT-SVID token and returns an instance of a {@link JwtSvid} with a SPIFFE ID parsed from the 'sub', audience from 'aud',
```
maxlambrecht, line 34: nit: `validates a JWT-SVID token`
also: don't know if `{@link }` works here, it looks like Javadoc syntax

test/svid/jwtsvid/pyjwttest.py
```diff
@@ -0,0 +1,49 @@
+from calendar import timegm
+import datetime
+import jwt
+from jwt import PyJWKClient, PyJWS
+
+token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImtpZCI6Ik5FRTFRVVJCT1RNNE16STVSa0ZETlRZeE9UVTFNRGcyT0Rnd1EwVXpNVGsxUWpZeVJrUkZRdyJ9.eyJpc3MiOiJodHRwczovL2Rldi04N2V2eDlydS5hdXRoMC5jb20vIiwic3ViIjoiYVc0Q2NhNzl4UmVMV1V6MGFFMkg2a0QwTzNjWEJWdENAY2xpZW50cyIsImF1ZCI6Imh0dHBzOi8vZXhwZW5zZXMtYXBpIiwiaWF0IjoxNTcyMDA2OTU0LCJleHAiOjE1NzIwMDY5NjQsImF6cCI6ImFXNENjYTc5eFJlTFdVejBhRTJINmtEME8zY1hCVnRDIiwiZ3R5IjoiY2xpZW50LWNyZWRlbnRpYWxzIn0.PUxE7xn52aTCohGiWoSdMBZGiYAHwE5FYie0Y1qUT68IHSTXwXVd6hn02HTah6epvHHVKA2FqcFZ4GGv5VTHEvYpeggiiZMgbxFrmTEY0csL6VNkX1eaJGcuehwQCRBKRLL3zKmA5IKGy5GeUnIbpPHLHDxr-GXvgFzsdsyWlVQvPX2xjeaQ217r2PtxDeqjlf66UYl6oY6AqNS8DH3iryCvIfCcybRZkc_hdy-6ZMoKT6Piijvk_aXdm7-QQqKJFHLuEqrVSOuBqqiNfVrG27QzAPuPOxvfXTVLXL2jek5meH6n-VWgrBdoMFH93QEszEDowDAEhQPHVs0xj7SIzA"
+kid = "NEE1QURBOTM4MzI5RkFDNTYxOTU1MDg2ODgwQ0UzMTk1QjYyRkRFQw"
+url = "https://dev-87evx9ru.auth0.com/.well-known/jwks.json"
```
maxlambrecht, line 8: Can we add this `jwks.json` as part of tests resources and pass a file uri to the `PyJWKClient` ?

scotte:
I'm not sure if the docstrings will parse correctly as they don't exactly seem to follow what I think is the reference at https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html though I'm not 100% sure. Specifically the indentation doesn't seem to be aligned following the examples here.

scotte:
The test JWTs are kind of hard to reason about what is being tested, how about adding a comment with the decoded JWT? That will help with debugging as well if tests start failing because of an expiration date, or something like that.

test/svid/jwtsvid/pyjwttest.py
```diff
@@ -0,0 +1,49 @@
...
+jwks_client = PyJWKClient(url)
+signing_key = jwks_client.get_signing_key_from_jwt(token)
+data = jwt.decode(
+    token,
+    signing_key.key,
+    algorithms=["RS256"],
```
scotte, line 14: What about the other algorithms that are supported by SPIFFE JWTs? Is that something we need to test? We should probably test an invalid algorithm as well. It looks like that validation isn't implemented yet (or maybe I just missed it), so ignore if this is already planned. https://github.com/spiffe/spiffe/blob/master/standards/JWT-SVID.md#21-algorithm

src/pyspiffe/svid/jwt_svid.py
glaucimar: Thanks Max, good catch!

src/pyspiffe/svid/jwt_svid.py
glaucimar: Ah! true... I need to check the link.. I will remove for now. Thanks :)

test/svid/jwtsvid/pyjwttest.py
glaucimar: I think we can and that is a good idea indeed. It will make it easier to read the test.

test/svid/jwtsvid/pyjwttest.py
glaucimar: I should not have commit this file just yet.. this is an easy path for me to make some tests on PyJWT. With that said, yes, we should try the other algorithms but this is required by the parse_validate method that I have not yet finished. I will make sure this is validated and tested, thanks!

src/pyspiffe/svid/jwt_svid.py
glaucimar: Thanks Max. Done

glaucimar:
> The test JWTs are kind of hard to reason about what is being tested, how about adding a comment with the decoded JWT? That will help with debugging as well if tests start failing because of an expiration date, or something like that.



> The test JWTs are kind of hard to reason about what is being tested, how about adding a comment with the decoded JWT? That will help with debugging as well if tests start failing because of an expiration date, or something like that.

glaucimar:
> I'm not sure if the docstrings will parse correctly as they don't exactly seem to follow what I think is the reference at https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html though I'm not 100% sure. Specifically the indentation doesn't seem to be aligned following the examples here.

Thanks for the pointer, Scott! I will check and validate the doc string as well.

test/svid/jwtsvid/pyjwttest.py
glaucimar: Removed the file.

test/svid/jwtsvid/test_jwt_svid.py
```diff
@@ -0,0 +1,278 @@
...
+    'test_input_token,test_input_audience, expected',
+    [
+        (
+            "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJzcGlmZmU6Ly90ZXN0Lm9yZy8iLCJuYW1lIjoiSm9obiBEb2UiLCJpYXQiOjE1MTYyMzkwMjIsImV4cCI6NDcyMTUwNTQzNywiYXVkIjpbInNwaXJlIl19.cvoi48geyYPAP88RQbBXyEXoRX09II5xAR0XKULWDwQ",
+            ["spire"],
+            "spiffe://test.org/",
+        ),
+    ],
+)
+def test_valid_parse_insecure(test_input_token, test_input_audience, expected):
```
maxlambrecht, line 267: add another test case, with a different token and multiple audiences

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,156 @@
...
+    def _validate_exp(cls, expiration_date: str) -> None:
+        expiration_date = int(expiration_date)
+        utctime = timegm(datetime.datetime.utcnow().utctimetuple())
+        if expiration_date < utctime:
+            raise JwtSvidError(SIGNATURE_EXPIRED_ERROR)
+
+    ###
+    ### Verifies if any item specified by audience is present in audience_clains.
+    ### Raises:
+    ###     JwtSvidError: in case none of the items specified by audience is present in audience_claims.
```
maxlambrecht, line 125: the error should be raised if the expected audience is not a subset of the audience in the token

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,156 @@
...
+    ###
+    @classmethod
+    def _validate_aud(cls, audience_claims: [], audience: []) -> None:
```
maxlambrecht, line 128: I'd rename `audience` as `expected_audience` and `audience_claims` as `audience_claim`

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,156 @@
...
+        if not audience_claims and not audience:
```
maxlambrecht, line 129: replace by `if not audience` to make it more clear. In case the `audience_claim` is empty it will not pass the validation of being a superset of the expected audience.

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,156 @@
...
-        self.privateKey = private_key
-        self.chain = chain
+    ###
+    ### Validates payload for required claims (aud, exp, sub), signature expiration and audience.
+    ###
+    ### Raises:
+    ###     JwtSvidError: in case a required claim is not present in payload, token is expired or aud claimm does not match audience parameter.
+    ###
+    @classmethod
+    def _validate_claims(cls, payload: {}, audience: []) -> None:
```
maxlambrecht, line 148: what is the difference between _validate_claims and _validate_aud ?

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,156 @@
...
+import datetime
+from calendar import timegm
+import jwt
+from jwt import exceptions
+
+EMPTY_TOKEN_ERROR = 'Token cannot be empty.'
+INVALID_INPUT_ERROR = 'Invalid input {}.'
+MISSING_X_ERROR = 'Token is missing {}.'
+AUDIENCE_NOT_MATCH_ERROR = 'Audience does not match payload[aud] in Token.'
+SIGNATURE_EXPIRED_ERROR = 'Token signature has expired.'
```
maxlambrecht, line 13: `TOKEN_EXPIRED_ERROR = 'Token has expired.'`

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,156 @@
...
+        self.token = token
 
-    :param spiffe_id: SPIFFE ID of the X509-SVID.
-    :param private_key: Audience is the intended recipients of JWT-SVID as present in the 'aud' claim.
-    :param chain: Expiration time of JWT-SVID as present in 'exp' claim.
     """
+    Parses and validates a JWT-SVID token and returns an instance of a JwtSvid with a SPIFFE ID parsed from the 'sub', audience from 'aud', 
+    and expiry from 'exp' claim. The JWT-SVID signature is not verified.
+    
+    Ags:
```
maxlambrecht, line 37: nit: `Args:`, also `Returns:` is missing

src/pyspiffe/svid/jwt_svid.py
glaucimar: _validate_claims validates all claims (aud, exp, sub), _validate_aud performs the validations for aud claim only.

src/pyspiffe/svid/jwt_svid.py
glaucimar: Done, thanks!

src/pyspiffe/svid/jwt_svid.py
glaucimar: Ok, done.

src/pyspiffe/svid/jwt_svid.py
glaucimar: In this case, the exception raised is different and I wanted it to be consistent: ValueError for empty input. But I am ok if this is going to be consistent with other spiffe libraries.. so, in case audience is specified and claim['aud'] does not exist or is empty, do the other libraries raise "Audience does not match kind of error"? I will check.. if that is the case, I can make the change as you suggest.

src/pyspiffe/svid/jwt_svid.py
glaucimar: That is a good suggestion, thanks! Done.

src/pyspiffe/svid/jwt_svid.py
glaucimar: Thanks Max, I will make the change to be consistent with what you described.

test/svid/jwtsvid/test_jwt_svid.py
glaucimar: I will do for _validate_claims not for parse_insecure, I was expecting the discussion about the right way to validate it. Thanks for the reminder.

src/pyspiffe/svid/jwt_svid_validator.py
```diff
@@ -0,0 +1,118 @@
...
+        if not header:
+            raise ValueError(INVALID_INPUT_ERROR.format('header cannot be empty'))
+
+        if header['alg'] not in self._supported_algorithms:
+            raise InvalidAlgorithmError(header['alg'])
+        try:
+            if header['typ'] and header['typ'] not in self._supported_types:
+                raise InvalidTypeError(header['typ'])
+        except KeyError:
+            pass
```
Andres-GC, line 62: Due to the check `if header['alg']` I assume that the dict `header` might not contain all the keys. In that case I think we need to used `dict.get(key, default_value)` here. It will return the value for the given `key` at the dictionary or a `default_value` which by default is `None`. So here `header.get('alg')` implicitly handles `KeyError` exception.

I think this needs to be rewritten as:
```
   alg = header.get('alg')
   if not alg:
       raise INVALID_INPUT_ERROR.format('header alg cannot be empty.')
   if alg not in self._supported_algorithms:
       raise InvalidAlgorithmError(alg)

   typ = header.get('typ')
   if not typ:
       raise INVALID_INPUT_ERROR.format('header type cannot be empty.')
       
    if typ not in self._supported_types:
        raise InvalidTypeError(typ)
```
Another option could be to just raise the error and not make any difference between an invalid and an empty alg/typ.

test/svid/jwtsvid/test_jwt_svid.py
```diff
@@ -0,0 +1,160 @@
...
+                        (
+                            datetime.datetime.utcnow() + datetime.timedelta(hours=72)
+                        ).utctimetuple()
+                    ),
+                },
+                'secret',
+                headers={'alg': 'RS256', 'typ': 'JOSE'},
+            ),
+            ["spire"],
+            InvalidClaimError._text.format('aud'),
```
Andres-GC, line 54: nit: Doesn't work  `InvalidClaimError('aud').text` here?
Accessing the private property directly looks at least weird to me.

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,118 @@
...
+
+    @classmethod
+    def parse_insecure(cls, token: str, expected_audience: []) -> 'JwtSvid':
+        if not token:
+            raise ValueError(INVALID_INPUT_ERROR.format('token cannot be empty'))
+        token_header = jwt.get_unverified_header(token)
+        validator = JwtSvidValidator()
+        validator.validate_header(token_header)
+        claims = jwt.decode(token, options={'verify_signature': False})
+        validator.validate_claims(claims, expected_audience)
```
Andres-GC, line 62: Not sure if is possible based on the uses cases but in the scenario that `jwt.any_method()` raises and exception, is it "excepted" at some point and translated to one of the "local exceptions" or is just raised?

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,118 @@
...
+        an instance of JwtSvid with a SPIFFE ID parsed from the 'sub', audience from 'aud', and expiry
+        from 'exp' claim.
+
+    Raises:
+        ValueError: when the token is blank or cannot be parsed, in case header is not specified,
+        InvalidAlgorithmError: in case specified 'alg' is not supported as specified by the SPIFFE standard.
+        InvalidTypeError: in case 'typ' is present in header but is not set to 'JWT' or 'JOSE'.
+        InvalidClaimError: in case a required claim is not present in payload or expected_audience is not a subset of audience_claim.
+        TokenExpiredError: in case token is expired.
+        ValueError: in case expected_audience is empty or if the string spiffe_id doesn't comply the SPIFFE standard.
```
maxlambrecht, line 51:
```
 ValueError: in case expected_audience is empty or if the SPIFFE ID in the 'sub' claim doesn't comply with the SPIFFE standard.
```

src/pyspiffe/svid/jwt_svid_validator.py
```diff
@@ -0,0 +1,118 @@
...
+    InvalidTypeError,
+)
+
+AUDIENCE_NOT_MATCH_ERROR = 'audience does not match expected value'
+INVALID_INPUT_ERROR = 'Invalid input: {}.'
+
+
+class JwtSvidValidator(object):
+    """Performs validations on a given token checking compliance to SPIFFE specification.
+    See <a href="https://github.com/spiffe/spiffe/blob/master/standards/JWT-SVID.md">https://github.com/spiffe/spiffe/blob/master/standards/JWT-SVID.md</a>.
```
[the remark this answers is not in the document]
maxlambrecht: I would use 
```
`SPIFFE JWT-SVID standard <https://github.com/spiffe/spiffe/blob/master/standards/JWT-SVID.md>`
```

test/svid/jwtsvid/test_jwt_svid.py
```diff
@@ -0,0 +1,160 @@
...
+                        (
+                            datetime.datetime.utcnow() + datetime.timedelta(hours=1)
+                        ).utctimetuple()
+                    ),
+                },
+                'secret key',
+                headers={'alg': 'PS384', 'typ': 'JOSE'},
+            ),
+            ['spire', 'test'],
+            "spiffe://test.orgcom.br/",
```
maxlambrecht, line 145: would be useful to add a testcase using a hardcoded token fetched from the workload api, like this one?

```
ES256TOKEN =  'eyJhbGciOiJFUzI1NiIsImtpZCI6Imd1eTdsOWZSQzhkQW1IUmFtaFpQbktRa3lId2FHQzR0IiwidHlwIjoiSldUIn0.eyJhdWQiOlsib3RoZXItc2VydmljZSJdLCJleHAiOjE2MTIyOTAxODMsImlhdCI6MTYxMjI4OTg4Mywic3ViIjoic3BpZmZlOi8vZXhhbXBsZS5vcmcvc2VydmljZSJ9.W7CLQvYVBQ8Zg3ELcuB1K9hE4I9wyCMB_8PJTZXbjnlMBcgd0VDbSm5OjoqcGQF975eaVl_AdkryJ_lzxsEQ4A'
```

maxlambrecht (commented):
Added a few more comments.

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,118 @@
...
+    """
```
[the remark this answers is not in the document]
glaucimar: Thanks Andres! Done!

src/pyspiffe/svid/jwt_svid_validator.py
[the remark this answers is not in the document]
glaucimar: Thank you! I've updated the docstring.

src/pyspiffe/svid/jwt_svid_validator.py
```diff
@@ -0,0 +1,118 @@
...
+
+from pyspiffe.svid.exceptions import (
+    TokenExpiredError,
+    InvalidClaimError,
+    InvalidAlgorithmError,
+    InvalidTypeError,
+)
+
+AUDIENCE_NOT_MATCH_ERROR = 'audience does not match expected value'
+INVALID_INPUT_ERROR = 'Invalid input: {}.'
```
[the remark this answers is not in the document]
glaucimar: Audience is used to complement text on InvalidClaimError. The text in this class is _text = 'Invalid claim value: {}.' So, the dot is there.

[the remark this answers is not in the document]
glaucimar: > `INVALID_INPUT_ERROR`: Is also defined at `jwt_svid.py` (the same way). Does it make sense to move the declaration to a package level and avoid duplication?

Yeap! it does... good catch.

src/pyspiffe/svid/exceptions.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __init__(self, text: str) -> None:
+        self.text = text
+
+    def __str__(self) -> str:
+        return self.text
+
+
+class InvalidClaimError(JwtSvidError):
+    """Raised when an erronic value is found in the JWT token claims (e.g  missing required claims, invalid claims values, etc).
```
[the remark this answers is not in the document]
glaucimar: I like erronic :) but changed anyway. :)

src/pyspiffe/svid/exceptions.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Args:
+        complement_text (str): complementary text to be appended to the default error message.
```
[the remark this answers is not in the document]
glaucimar: Well... it really complements the description of the Error adding to the default one. Maybe additional would be better?

src/pyspiffe/svid/exceptions.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class JwtSvidError(PySpiffeError):
+    """Top level exception for JwtSvid module.
+
+    Args:
+        text (str): message describing the error.
+
+    Attributes:
+        text (str): message describing the error. This is the message to be returned when __str__ is called.
```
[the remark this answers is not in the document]
glaucimar: Yeap, much better. Thanks Max!

src/pyspiffe/svid/exceptions.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class InvalidClaimError(JwtSvidError):
+    """Raised when an erronic value is found in the JWT token claims (e.g  missing required claims, invalid claims values, etc).
+
+    Args:
+        complement_text (str): complementary text to be appended to the default error message.
+
+    Attributes:
+        _text (str): error message describing the encountered error.
```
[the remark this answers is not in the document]
glaucimar: This is the same case as before, it adds information to the default message. About the underline, it is a python convention used to identify "private" methods or attributes. meaning that developer is informing you that you are not expected to use these. But this is just a convention as everything is public in Python.

I've changed _text to _MESSAGE as this is supposed to be static. And changed the input parameter to additional_information. Let me know if you still have comments there.

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,118 @@
...
 
 
 class JwtSvid(object):
     """
-    Represents a SPIFFE X.509 SVID.
+    Represents a SPIFFE JWT SVID as defined in the SPIFFE standard.
+    See <a href="https://github.com/spiffe/spiffe/blob/master/standards/JWT-SVID.md">https://github.com/spiffe/spiffe/blob/master/standards/JWT-SVID.md</a>.
+
+    Attributes:
+        spiffeId (SpiffeId): token spiffeId.
```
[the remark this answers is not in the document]
glaucimar: Changed, thanks for the improved text.

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,118 @@
...
+    Args:
+        token(str): a token as a string that is parsed and validated.
+        param audience(List): audience as a list of strings used to validate the 'aud' claim.
+
+    Returns:
+        an instance of JwtSvid with a SPIFFE ID parsed from the 'sub', audience from 'aud', and expiry
+        from 'exp' claim.
+
+    Raises:
+        ValueError: when the token is blank or cannot be parsed, in case header is not specified,
```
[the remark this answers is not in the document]
glaucimar: Changed, thanks Max!

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,118 @@
...
+        InvalidAlgorithmError: in case specified 'alg' is not supported as specified by the SPIFFE standard.
+        InvalidTypeError: in case 'typ' is present in header but is not set to 'JWT' or 'JOSE'.
+        InvalidClaimError: in case a required claim is not present in payload or expected_audience is not a subset of audience_claim.
```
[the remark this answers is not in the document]
glaucimar: Done!

src/pyspiffe/svid/jwt_svid.py
glaucimar: Ok, done.

test/svid/jwtsvid/test_jwt_svid.py
glaucimar: I think so, I will do it.

test/svid/jwtsvid/test_jwt_svid.py
glaucimar: That is a great catch, Andres! Thanks for raising it. I've fixed it.

src/pyspiffe/svid/jwt_svid.py
glaucimar: That is a good point, Andres! I will make the change to catch it and translate to a "local" error-exception.

src/pyspiffe/svid/jwt_svid_validator.py
glaucimar: Thanks Andres for the tip! I changed the code to use .get method. It is much better!

glaucimar: Ah! I also changed the other occurrences of the usage of Dict too. Thanks!

src/pyspiffe/svid/jwt_svid.py
glaucimar: Done!

test/svid/jwtsvid/test_jwt_svid.py
glaucimar: @maxlambrecht , as we chatted, SPIRE does not generated a long-very long lived token and after sometime, I will start getting TokenExpiredException when I was supposed to get a valid JwtSVID. I will put this test idea on hold for the future if we find a way around this. Thanks!

Pipfile
```diff
@@ -8,8 +8,9 @@ grpcio = "~=1.33.2"
 grpcio-tools = "~=1.33.2"
 requests = "~=2.25.0"
 python-json-logger = "~=0.1.11"
-cryptography = "~=3.2.1"
+cryptography = "~=3.3.1"
```
maxlambrecht, line 11: cryptography dependency has just been updated to `3.4.4`, rebase to pick that change.

maxlambrecht (approved):
LGTM! 🏅

src/pyspiffe/svid/jwt_svid_validator.py
```diff
@@ -0,0 +1,132 @@
...
+            raise ValueError(INVALID_INPUT_ERROR.format('header alg cannot be empty'))
+
+        if alg not in self._SUPPORTED_ALGORITHMS:
+            raise InvalidAlgorithmError(alg)
+
+        typ = header.get('typ')
+        if typ and typ not in self._SUPPORTED_TYPES:
+            raise InvalidTypeError(typ)
+
+    def validate_claims(self, payload: Dict, expected_audience: List) -> None:
```
Andres-GC, line 73: nit: `audience: List[str]` For `payload: Dict` you can also define the dict structure if it is fixed like `Dict[str, str]` for example.

src/pyspiffe/svid/jwt_svid_validator.py
```diff
@@ -0,0 +1,132 @@
...
+        Note: If and when https://github.com/jpadilla/pyjwt/issues/599 is fixed, this can be simplified/removed.
+
+        Args:
+            expiration_date: date to check if it is expired.
+
+        Raises:
+            TokenExpiredError: in case it is expired.
+            InvalidClaimError: in case expiration_date is not provided.
+        """
+        if not expiration_date or expiration_date == '':
```
Andres-GC, line 105: nit: `if not expiration_date` should already cover the scenario where `expiration_date == ''` because strings are [falsy](https://docs.python.org/3.6/library/stdtypes.html#truth-value-testing) in Python

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
```diff
@@ -1,2 +1,31 @@
...
+            TODO: complete
+
+        Args:
+            key_id: key of the token to return the correspondent bundle.
+
+        Returns:
+            TBD
+
+        Raises
+            JwtBundleNotFoundError  when no authority is found for the given key_id.
```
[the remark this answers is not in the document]
glaucimar: Done, thanks!

src/pyspiffe/svid/exceptions.py
```diff
@@ -0,0 +1,100 @@
...
+class TokenExpiredError(JwtSvidError):
+    """Raised when the JWT token is expired."""
+
+    _MESSAGE = 'Token has expired.'
+
+    def __init__(self) -> None:
+        """Creates an instance of TokenExpiredError
+
+        Args:
+            aadditional_information: additional information about the error.
```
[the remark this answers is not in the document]
glaucimar: Yeah.. in fact, there is no such argument in this exception :S... Copy and paste error. Fixed, thanks!

src/pyspiffe/svid/exceptions.py
```diff
@@ -0,0 +1,100 @@
...
+class InvalidAlgorithmError(JwtSvidError):
+    """Error raised when an invalid value is found in the JWT token's algorithm field."""
+
+    _MESSAGE = INVALID_VALUE_ERROR
+
+    def __init__(self, additional_information: str = '') -> None:
+        """Creates an instance of InvalidAlgorithmError
+
+        Args:
+            aadditional_information: additional information about the error.
```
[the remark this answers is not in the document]
glaucimar: Fixed, thanks!

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,140 @@
...
+from pyspiffe.bundle.jwt_bundle.jwt_bundle import JwtBundle
+from pyspiffe.svid.jwt_svid_validator import JwtSvidValidator
+from pyspiffe.svid.exceptions import InvalidTokenError
 
 
 class JwtSvid(object):
+    """Represents a SPIFFE JWT SVID as defined in the SPIFFE standard.
+    See `SPIFFE JWT-SVID standard <https://github.com/spiffe/spiffe/blob/master/standards/JWT-SVID.md>`
+
+    Attributes:
```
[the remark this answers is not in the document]
glaucimar: Perfect, done. Thanks!

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,140 @@
...
+            validator = JwtSvidValidator()
+            validator.validate_header(token_header)
+            claims = jwt.decode(token, options={'verify_signature': False})
+            validator.validate_claims(claims, expected_audience)
+            spiffe_ID = SpiffeId.parse(claims['sub'])
+            result = JwtSvid(spiffe_ID, claims['aud'], claims['exp'], claims, token)
+        except PyJWTError as err:
+            raise InvalidTokenError(err)
+
+        return result
```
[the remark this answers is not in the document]
glaucimar: It does make a lot of sense. Thanks, Andres!

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,140 @@
...
+
+    @classmethod
+    def parse_and_validate(
+        cls, token: str, jwt_bundle: JwtBundle, audience: List
```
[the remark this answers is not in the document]
glaucimar: Done, thanks!

src/pyspiffe/svid/jwt_svid.py
```diff
@@ -1,19 +1,140 @@
...
+            raise InvalidTokenError(err)
+
+        spiffe_ID = SpiffeId.parse(claims['sub'])
+        result = JwtSvid(spiffe_ID, claims['aud'], claims['exp'], claims, token)
 
-    def __init__(self, spiffe_id: SpiffeId, private_key: str, chain: List[str]):
-        self.spiffe_id = spiffe_id
-        self.privateKey = private_key
-        self.chain = chain
+        return result
```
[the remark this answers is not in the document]
glaucimar: Done.

src/pyspiffe/svid/jwt_svid_validator.py
glaucimar: The payload is not fixed so, it does not apply. But header is, so, I've changed for header.

src/pyspiffe/svid/jwt_svid_validator.py
glaucimar: Done.

Andres-GC (dismissed):
Looks great :medal_military: 
Thanks Glaucimar for your patience.

maxlambrecht (dismissed):
LGTM! 🥇

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
```diff
@@ -1,2 +1,33 @@
...
+        """Returns the authority for the specified key_id.
+            TODO: complete
+
+        Args:
+            key_id: key of the token to return the correspondent bundle.
+
+        Returns:
+            TBD
+
+        Raises
```
scotte, line 25: nit: missing `:`

src/pyspiffe/svid/exceptions.py
```diff
@@ -0,0 +1,95 @@
...
+INVALID_VALUE_ERROR = '{} is not supported.'
+"""str: not supported error message."""
+
+
+class JwtSvidError(PySpiffeError):
+    """Top level exception for JwtSvid module.
+
+    Attributes:
+        message: message describing the error.
+
```
scotte, line 16: nit: Extra newline (a few instances of this)

scotte (dismissed):
Just a couple of nits on the docstrings. Looks good!

Andres-GC (approved):

src/pyspiffe/svid/exceptions.py
glaucimar: Done! Thanks.

scotte (approved):

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
glaucimar: Nice! Thanks.

PR #29, opened 2021-01-21
Config setter improvements

mchurichi:
- Adds the ability to set the socket endpoint programmatically
- Returns a Config object instead of a dict

Usage:
```python
config = ConfigSetter(spiffe_endpoint_socket='unix:///path/to/endpoint.sock')
config.spiffe_endpoint_socket
```

`ConfigSetter` first reads from the environment variable `SPIFFE_ENDPOINT_SOCKET`, if set, and then overrides the value with the parameter `spiffe_endpoint_socket`, if set. It raises a `ValueError` when neither of the values are set, or if it's set but is has an invalid socket path.

### DIFF
--- a/src/pyspiffe/config.py
+++ b/src/pyspiffe/config.py
@@ -71,7 +71,7 @@ class ConfigSetter:
     def _validate_tcp_socket(cls, socket: ParseResult) -> None:
         try:
             ipaddress.ip_address(socket.hostname)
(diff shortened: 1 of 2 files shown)


scotte (approved):
Thank you Max!

Andres-GC (approved):
Thanks Max

Issue #13, opened 2020-12-17
Implement Workload API Default client 


Andres-GC:
This issue is to implement a primitive wrapper around the Workload API

For the most part, the implementation should more or less be a wrapper around the Workload API gRPC client, converting to/from protobufs. All of the methods on the client are synchronous and straightforward.

Depends on [#6](https://github.com/HewlettPackard/py-spiffe/issues/6)

scotte:
Depends on #6

scotte:
https://github.com/HewlettPackard/py-spiffe/pull/31

PR #31, opened 2021-01-22
Add initial workloadapi stub

scotte:
Initial stub of workload API client - Starts #13

### DIFF
--- a/src/pyspiffe/workloadapi/default_workload_api_client.py
+++ b/src/pyspiffe/workloadapi/default_workload_api_client.py
@@ -1,0 +1,100 @@
+from pyspiffe.bundle.x509_bundle.x509_bundle_set import X509BundleSet
+from pyspiffe.bundle.jwt_bundle.jwt_bundle_set import JwtBundleSet
+from pyspiffe.config import ConfigSetter
+from pyspiffe.svid.x509_svid import X509Svid
+from pyspiffe.svid.jwt_svid import JwtSvid
+
+from pyspiffe.workloadapi.workload_api_client import WorkloadApiClient
+
+
+class DefaultWorkloadApiClient(WorkloadApiClient):
+    """Default implementation for a SPIFFE Workload API Client."""
+
+    def __init__(self, spiffe_socket: str = None) -> None:
+        """Creates a new Workload API Client.
+
+        Args:
+            spiffe_socket (str, optional): Path to Workload API UDS. If
+                not specified, the SPIFFE_ENDPOINT_SOCKET environment variable
+                must be set.
+
+        Returns:
+            DefaultWorkloadApiClient: New Workload API Client object.
+
+        Raises:
+            ValueError: If spiffe_socket_path is invalid or not provided and
+                SPIFFE_ENDPOINT_SOCKET environment variable doesn't exist
+        """
+
+        try:
+            self._config = ConfigSetter(
+                spiffe_endpoint_socket=spiffe_socket
+            ).get_config()
+        except ValueError:
+            raise ValueError(
+                'SPIFFE socket argument or environment variable invalid in DefaultWorkloadApiClient.'
+            )
+
+    def get_spiffe_endpoint_socket(self) -> str:
+        """Returns the spiffe endpoint socket config for this WorkloadApiClient.
+
+        Returns:
+            str: spiffe endpoint socket configuration value
+        """
+
+        return self._config.spiffe_endpoint_socket
+
+    def fetch_x509_svid(self) -> X509Svid:
+        """Fetches a SPIFFE X.509-SVID.
+
+        Returns:
+            X509Svid: Instance of X509Svid object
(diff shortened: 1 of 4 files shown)


src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -0,0 +1,109 @@
...
+        try:
+            self._config = ConfigSetter(
+                spiffe_endpoint_socket=spiffe_socket
+            ).get_config()
+        except ValueError:
+            raise ValueError(
+                'SPIFFE socket argument or environment variable invalid in DefaultWorkloadApiClient'
+            )
+
+    def get_config(self) -> Config:
```
maxlambrecht, line 41: It's probably not such a good idea to expose the Config object, what would the use of this `get` method ?
In case you decide to keep it, use `config(self)`

src/pyspiffe/workloadapi/workload_api_client.py
```diff
@@ -0,0 +1,66 @@
+from abc import ABC, abstractmethod
+from pyspiffe.bundle.x509_bundle.x509_bundle_set import X509BundleSet
+from pyspiffe.bundle.jwt_bundle.jwt_bundle_set import JwtBundleSet
+from pyspiffe.svid.x509_svid import X509Svid
+from pyspiffe.svid.jwt_svid import JwtSvid
+
+
+class WorkloadApiClient(ABC):
```
maxlambrecht, line 8: 1) Document the methods `fetch_` describing that they are a `one-shot` call. 
2) The `watch_` methods are missing: e.g. `watch_x509_svids`, `watch_jwt_bundles`

maxlambrecht (commented):
It's looking good, just left a couple of comments.

src/pyspiffe/workloadapi/workload_api_client.py
scotte: The `watch` methods are missing be design since we are only implementing synchronous calls in py-spiffe at this time. I didn't think adding the `one-shot` text improved clarity as those are the only call forms for now. I'm happy to add (1) in, but I think we should leave (2) off since async calls are not currently scoped for the library.

src/pyspiffe/workloadapi/default_workload_api_client.py
scotte: Resolved in latest commit, with get_config() replaced by get_spiffe_endpoint_socket()

mchurichi:
It looks good to me, I just have a few minor comments about the docstrings usage and coding style.

Based on the [google styleguide](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) we are following and the [example file](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html), there are a few tweaks we can make here:

- The summary line should be in the same line where the docstring quotes start:
 eg:
```python
def __init__(self, spiffe_socket: str = None):
    """Creates a new Workload API Client.
```
- The summary should be in descriptive-style instead of imperative-style. E.g. `Returns` vs `Return`
- The summary line should end with period, question or exclamation mark.
- Args' parameters types are not mandatory when using typings in the function's signature. Eg: `spiffe_socket: Path to Workload API UDS.`
- `Raises`' exception names should be followed by a colon: E.g.: `ValueError: if spiffe_socket_path is invalid...`

There are also two additional rules (@Andres-GC will share more details during tomorrow's py-spiffe sync call):

- Exception messages should end with a period. E.g: `raise ValueError('SPIFFE socket argument or environment variable invalid in DefaultWorkloadApiClient.')`
- `__init__` functions should be marked with return `None`. Eg: `def __init__(self, spiffe_socket: str = None) -> None:`

The last one looks a bit tricky, but it's the right way to annotate it, and we believe that typing every function we have would make our codebase more consistent.

scotte:
@mchurichi I will go through the docstrings and try to make them conform to the Google style guide, even though it's less readable in some ways, unfortunately.

mchurichi:
@scotte yes, I agree some rules don't make the docstrings easier to consume for humans. Anyways, I don't want to bug anyone with minor rules especially when we don't have all of them 100% defined. @Andres-GC is working in some guidelines on top of Google Style we should discuss before starting to apply them, so feel free to ignore my comments in the meantime, we can revisit this code later on if needed.

mchurichi (dismissed):
LGTM :shipit:

Andres-GC (dismissed):
:ice_cream:

maxlambrecht (dismissed):

scotte:
Updated docstrings per @mchurichi  comments.

scotte:
And...once more to rebase on master.

Andres-GC (approved):
Re-Approved :medal_military:

Andres-GC (approved):

mchurichi (approved):
LGTM. Thanks for this, Scott

PR #32, opened 2021-01-25
Remove option java_package from protobuf file.

maxlambrecht:
We are not using the option `java_package` from the protobuf definition, so better remove it.
 

proto/workload.proto
```diff
@@ -1,7 +1,5 @@
 syntax = "proto3";
 
-option java_package = "io.spiffe.workloadapi.grpc";
```
Andres-GC, line 3: Good Catch Max. 
This doesn't require the *_pb files to be regenerated?
If it is not necessary but resulting files are different, them I'm in favor of generate the new ones and update the code.

maxlambrecht: Sorry, missed that. It does change, doesn't affect anything but still worthwhile to keep them in sync. 
Thanks for the catch.

Andres-GC (approved):
Great Max :medal_military:

mchurichi (approved):
LGTM

PR #33, opened 2021-01-26
More tests and docstrings improvements for the config setter class

mchurichi:
Signed-off-by: Maximiliano Churichi <maximiliano.churichi@hpe.com>

### DIFF
--- a/src/pyspiffe/config.py
+++ b/src/pyspiffe/config.py
@@ -1,70 +1,84 @@
+"""Module that contains Configuration related classes
+"""
+
 import os
 import ipaddress
 from urllib.parse import ParseResult, urlparse
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
 
-SPIFFE_ENDPOINT_SOCKET = 'SPIFFE_ENDPOINT_SOCKET'
+_SPIFFE_ENDPOINT_SOCKET = 'SPIFFE_ENDPOINT_SOCKET'
 
 
 class Config:
-    """
-    Represents the configuration for a Workload API client
+    """Represents the configuration for a Workload API client.
+
+    Attributes:
+        spiffe_endpoint_socket (str): Path to the Workload API UDS.
     """
 
-    def __init__(self, spiffe_endpoint_socket: str):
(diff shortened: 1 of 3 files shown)


test/test_config.py
```diff
@@ -30,7 +30,15 @@ def test_read_socket_from_environment_variables():
     os.environ['SPIFFE_ENDPOINT_SOCKET'] = fake_socket
 
     setter = ConfigSetter()
-    del os.environ['SPIFFE_ENDPOINT_SOCKET']
```
scotte, line 33: Does the environment variable still get deleted at the end of testing? Just want to make sure environment state is reset at the end is all. Thanks!

scotte (approved):
I have just one question about making sure the env is clean at the end, otherwise looks good to me.

test/test_config.py
mchurichi: yes, there's a post step for this suite that restores the environment state after each test run: https://github.com/HewlettPackard/py-spiffe/pull/33/files#diff-45fa5f8530eb815c0cf606032e587adedb04dc70e59b203535d51c44654cfc4dR7

mchurichi: now that you mention this, I think this is a nice util function that we can reuse across test suites

scotte: Ah thanks, that makes sense why I didn't see it now. Thanks!

glaucimar (approved):
Looks good to me.

src/pyspiffe/config.py
```diff
@@ -8,17 +8,23 @@
 
 
 class Config:
-    """
-    Represents the configuration for a Workload API client
+    """Represents the configuration for a Workload API client
```
Andres-GC, line 11: nit: _The summary line should end with period, question or exclamation mark._ Both Summary lines in this file do not have any of those ending characters.

mchurichi: good catch!

Andres-GC (approved):
:medal_military:

glaucimar (approved):

scotte (approved):
Re-approve

PR #36, opened 2021-02-09
Update mypy task to run against the src package.

Andres-GC:
`--ignore-missing-imports` option is used to avoid issues related to third party libraries missing type hints. In our code, this is the case for `rfc3987`. [Docs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-ignore-missing-imports)

### DIFF
--- a/src/pyspiffe/spiffe_id/trust_domain.py
+++ b/src/pyspiffe/spiffe_id/trust_domain.py
@@ -27,7 +27,7 @@ class TrustDomain(object):
         spiffe://domain.test
     """
 
-    def __init__(self, name: str):
+    def __init__(self, name: str) -> None:
         self.__set_name(name)
 
     def __str__(self) -> str:
--- a/src/pyspiffe/spiffe_id/trust_domain.py
+++ b/src/pyspiffe/spiffe_id/trust_domain.py
@@ -1,5 +1,5 @@
-from urllib.parse import urlparse
-from typing import Tuple, Any
+from urllib.parse import urlparse, ParseResult
+from typing import Any, cast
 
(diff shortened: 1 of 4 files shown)


scotte (approved):
Thanks!

src/pyspiffe/spiffe_id/trust_domain.py
```diff
@@ -56,9 +56,9 @@ def __set_name(self, name: str) -> None:
             )
 
         name = self.normalize(name)
-        uri = urlparse(name)
+        uri: ParseResult = urlparse(name)
         self.validate_uri(uri)
-        self.__name = uri.hostname.lower().strip()
+        self.__name = cast(str, uri.hostname).lower().strip()
```
Andres-GC, line 61: The error here was: `error: Item "None" of "Optional[str]" has no attribute "lower"`
`self.validate_uri(uri)` already checks if `hostname` is None and raise an exception in that case. Mypy is no able to detect that so it complains about `hostname` been `None`.
A way to solve that is using `Typing.cast(type, value)`.
In other places it could make sense to directly have something like:
```
        uri: ParseResult = urlparse(name)
        if not uri.hostname:
            raise ValueError(' ')
        self.__name = uri.hostname.lower().strip()
```
And in that case Mypy won't complain.

scotte: Thanks for the explanation, I was a bit confused why mypy was complaining about this.

scotte (approved):

glaucimar (approved):

PR #37, opened 2021-02-10
Updates cryptography version due to a Security Issue

Andres-GC:
There is a new version of the cryptography library that includes a fix for a security issue: 
```
3.3.2 - 2021-02-07
SECURITY ISSUE: Fixed a bug where certain sequences of update() calls when symmetrically encrypting very 
large payloads (>2GB) could result in an integer overflow, leading to buffer overflows. CVE-2020-36242
```
[Source](https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#332---2021-02-07)

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -8,7 +8,7 @@ grpcio = "~=1.33.2"
 grpcio-tools = "~=1.33.2"
 requests = "~=2.25.0"
 python-json-logger = "~=0.1.11"
-cryptography = "~=3.2.1"
+cryptography = "~=3.4.4"
 rfc3987 = "~=1.3.8"
 
 [requires]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "198792c65de45e01743c0b9393e589d0c8c2fc42aa871531705d8b9eea3722fe"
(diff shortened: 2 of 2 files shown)


Pipfile
```diff
@@ -8,7 +8,7 @@ grpcio = "~=1.33.2"
 grpcio-tools = "~=1.33.2"
 requests = "~=2.25.0"
 python-json-logger = "~=0.1.11"
-cryptography = "~=3.2.1"
+cryptography = "~=3.3.2"
```
maxlambrecht, line 11: why not the latest version so far: `3.4.4` ?

Andres-GC: I know that you have work in progress that use this library and also there is a PR from Glaucimar that use it. I just didn't want to have a bigger change to avoid possible issues, If you and Glaucimar are ok with the latest version, then I'm happy to update it.

maxlambrecht: we are all using `3.x`,  they should be compatible, so I'm ok with the latest.

maxlambrecht: I tested my WIP with `3.4.4`, it works.

glaucimar: I also tested my WIP with 3.4.4 and it worked too.

Andres-GC: Great, thanks to both. I'll update it to the latest version then.

glaucimar (approved):

maxlambrecht (approved):
💯

scotte (approved):
Let's do it!

PR #40, opened 2021-02-11
Moves proto files inside src directory to include them into the library

Andres-GC:
Besides moving `proto` directory into `src` it also includes the following changes:
- Adds new directory `spiffe` inside proto, This is something I noted on go-spiffe and I liked it.
- Updates make tasks `lint` and `pb_generate` according to the new location of the proto files.

**UPDATE**

I updated the PR with two other changes:
- Includes `--recreate` option for `tox` to force the re-creation of the virtual env where the library is installed before running the tests. (This helps when there are new dependencies that needs to be added to setup.py)
- Update `mypy` task. This is far away from nice but it is a workaround unitl the new version of mypy is released. [It should be in a couple of days](https://github.com/python/mypy/issues/10062)


Makefile
```diff
@@ -57,11 +57,11 @@ docs:
 pb_generate:
 	@echo "Generates pb files."
 	$(PIPENV_CMD) run python -m grpc_tools.protoc \
-		--proto_path=./proto \
-		--mypy_out=./proto \
+		--proto_path=./src/pyspiffe/proto/spiffe \
+		--mypy_out=./src/pyspiffe/proto/spiffe \
 		--proto_path=. \
```
glaucimar, line 62: Is the second "--proto_path=..." really needed? I executed a test without it and could generate the python files successfully.

scotte: Good catch, that is definitely not needed.

scotte (commented):
All looks great except for Glaucimar's catch. Thanks!

Makefile
Andres-GC: Yes, is not necessary. And while checking that I found I need to make another change so thanks for pointing it out.

scotte (approved):
Excellent!

maxlambrecht (approved):

mchurichi (approved):
LGTM

Issue #17, opened 2020-12-18
Implement x509_svid.SVID type


Andres-GC:
x509_svid.SVID represents an X509-SVID with the accompanying private key. Most of the accompanying methods are for marshaling/unmarshaling. It also implements the x509_svid.Source interface.

PR #41, opened 2021-02-16
Implementation of X.509 SVID type.

maxlambrecht:
This PR implements the SPIFFE X.509-SVID type and the utility methods to parse X509Svid instances from PEM and DER encoded bytes.

Solves #17 


glaucimar (commented):
HI Max, I've added some small comments.

src/pyspiffe/svid/exceptions.py
```diff
@@ -90,3 +90,135 @@ def __init__(self, additional_information: str = '') -> None:
             additional_information: additional information about the error.
         """
         super().__init__(self._MESSAGE.format(additional_information))
+
+
+class X509SvidError(PySpiffeError):
+    """Top level exception for the X509Svid module.
+
+    Attributes:
+        message: message describing the error.
```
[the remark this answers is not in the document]
maxlambrecht: the rest of the codebase uses lowercase after the colon in the params descriptions

src/pyspiffe/svid/exceptions.py
```diff
@@ -90,3 +90,135 @@ def __init__(self, additional_information: str = '') -> None:
...
+        self.message = message
+
+    def __str__(self) -> str:
+        return self.message
+
+
+class InvalidLeafCertificateError(X509SvidError):
+    """Error raised when an invalid leaf certificate is found in the X.509 chain."""
+
+    _MESSAGE = 'Invalid leaf certificate: {}'
```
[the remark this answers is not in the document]
maxlambrecht: Refactored adding the ending dots to the exception messages in the exceptions classes.

src/pyspiffe/svid/exceptions.py
```diff
@@ -90,3 +90,135 @@ def __init__(self, additional_information: str = '') -> None:
...
+
+    def __init__(self, additional_information: str = '') -> None:
+        """Creates an instance of InvalidLeafCertificateError
+
+        Args:
+            additional_information: additional information about the error.
```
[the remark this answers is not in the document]
maxlambrecht: well, it's not really optional, if it is not provided the message is not well formatted.

test/svid/x509svid/test_x509_svid.py
```diff
@@ -0,0 +1,389 @@
...
+    LoadPrivateKeyError,
+)
+from pyspiffe.svid.x509_svid import X509Svid
+from cryptography.hazmat.primitives.asymmetric import ec, rsa
+
+_TEST_CERTS_PATH = 'test/svid/x509svid/certs/{}'
+
+
+def test_parse_raw_chain_and_ec_key():
+    chain_bytes = read_bytes(_TEST_CERTS_PATH.format('1-chain.der'))
```
Andres-GC, line 21: I see that `read_bytes` is a helper function defined here for the tests. Is there any reason that prevents from setting the path internally in the function? That way `_TEST_CERTS_PATH.format('file_name')` occurs only once.

src/pyspiffe/svid/x509_svid.py
```diff
@@ -1,34 +1,475 @@
...
+        except FileNotFoundError:
+            raise LoadPrivateKeyError(
+                'Private key file not found: {}'.format(private_key_path)
+            )
+        except Exception as err:
+            raise LoadPrivateKeyError(
+                'Private key file could not be read: {}'.format(str(err))
+            )
+
+    @staticmethod
```
Andres-GC, line 286: To me, looks like all static methods here can be moved from the X509Svid object to simple functions in a different module.

src/pyspiffe/svid/x509_svid.py
```diff
@@ -1,34 +1,475 @@
...
+            private_key: a Private Key object.
+        """
+
+        self._spiffe_id = spiffe_id
+        self._cert_chain = cert_chain
+        self._private_key = private_key
+
+    def leaf(self) -> Certificate:
+        """Returns the Leaf X.509 certificate of the chain."""
+        return self._cert_chain[0]
```
Andres-GC, line 76: Is it safe to call `cert_chain[0]` without validating that there is at least on cert in the chain?

src/pyspiffe/svid/x509_svid.py
maxlambrecht: do you think? These static methods are very specific to the X509Svid type and I don't see any possible use for other modules.

src/pyspiffe/svid/x509_svid.py
```diff
@@ -1,34 +1,475 @@
...
+        if encoding == serialization.Encoding.PEM:
+            return cls.parse(chain_bytes, key_bytes)
+        elif encoding == serialization.Encoding.DER:
+            return cls.parse_raw(chain_bytes, key_bytes)
+        else:
+            raise ValueError(
+                'Encoding not supported: {}. Expected \'PEM\' or \'DER\'.'.format(
+                    encoding
+                )
+            )
```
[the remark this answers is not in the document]
maxlambrecht: Done

src/pyspiffe/svid/x509_svid.py
```diff
@@ -1,34 +1,475 @@
...
+            StoreCertificateError: in the case the file path in certs_chain_path cannot be open to write,
+                                  or there is an error storing the certificates to the file.
         """
-        self.spiffe_id = spiffe_id
-        self.audience = audience
-        self.expiry = expiry
-        self.claims = claims
-        self.token = token
+
+        if not (encoding is encoding.PEM or encoding is encoding.DER):
```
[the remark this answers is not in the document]
maxlambrecht: Good catch.

test/svid/x509svid/test_x509_svid.py
maxlambrecht: Done.

src/pyspiffe/svid/x509_svid.py
maxlambrecht: It is safe because the methods that create an X509Svid enforce that the cert_chain have at least one Certificate, we can't have an X509Svid with an empty chain.

maxlambrecht: The X509Svid constructor now enforce the invariants.

src/pyspiffe/svid/x509_svid.py
maxlambrecht: ok, I refactored those methods as simple functions in the X509Svid module, prefixing them with an underscore for suggesting that they are internal and not meant to be used outside of the library (as an implementation detail they can change later on without breaking the interface contract).

scotte (approved):

src/pyspiffe/svid/x509_svid.py
```diff
@@ -1,34 +1,472 @@
...
+            x509_cert = x509.load_pem_x509_certificate(
+                cert.as_bytes(), default_backend()
+            )
+            result.append(x509_cert)
+        except Exception:
+            raise ParseCertificateError(_UNABLE_TO_LOAD_CERTIFICATE)
+
+    if len(result) < 1:
+        raise ParseCertificateError(_UNABLE_TO_LOAD_CERTIFICATE)
+    return result
```
Andres-GC, line 414: If I got this correct, if there is a cert that cannot be parse, then it will raise an ParseCertificateError. On the other hand, if there is no cert at all, it will raise an ParseCertificateError too.

Does it makes sense to change this a little bit like this:
```
def _parse_pem_certificates(chain_pem_bytes: bytes) -> List[Certificate]:
    parsed_certs = pem.parse(chain_pem_bytes)
    if not parsed_certs:  # If not certs to parse, then raise error.
            raise ParseCertificateError(_UNABLE_TO_LOAD_CERTIFICATE)

    result = []
    for cert in parsed_certs:
        try:
            x509_cert = x509.load_pem_x509_certificate(
                cert.as_bytes(), default_backend()
            )
            result.append(x509_cert)
        except Exception:
            raise ParseCertificateError(_UNABLE_TO_LOAD_CERTIFICATE)

    return result # At this point at least one cert is parsed. There is no need to check result.
```

src/pyspiffe/svid/x509_svid.py
```diff
@@ -1,34 +1,472 @@
...
+        encoding: serialization.Encoding,
+    ) -> 'X509Svid':
+        """Loads the X.509 SVID from PEM or DER encoded files on disk.
+
+        The private key should be without encryption.
+
+        Args:
+            certs_chain_path (str): Path to the file containing one or more X.509 certificates as PEM blocks.
+            private_key_path (str): Path the file containing a private key as PKCS#8 PEM block.
+            encoding (serialization.Encoding): The encoding used to serialize the certs and private key, can be
```
Andres-GC, line 193: Type information should not be part of the docstring here. This is already included on the arguments declaration.
Save method also needs to be update to remove the typing from the docstring.

Andres-GC (commented):
Looks great Max. I've only commented 3 minor things. Please check if my comments are correct.

src/pyspiffe/svid/x509_svid.py
maxlambrecht: that is correct and more clear, I made the change.

Andres-GC (approved):

glaucimar (approved):
LGTM

PR #43, opened 2021-02-16
Fixes #42 - Pass --recreate option to tox

### LINKED ISSUE #42: "make test" runs tox without --recreate, cached files might be out of date
Unless `tox` is run with the `--recreate` option, it will used cached dependencies that might be out of date. `make test` should pass this option. This does result in tests taking a bit longer to run as the venv and dependencies have to be created and copied.

scotte:
Without this option, "make test" uses a cached venv with
dependencies and packages that might be out of date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ dev: env
 ## Runs unit tests.
 test: lint
 	@echo "Running unit tests."
-	$(PIPENV_CMD) run tox
+	$(PIPENV_CMD) run tox --recreate
 
 
 ## Lint source files.


maxlambrecht (approved):
💯

glaucimar:
I wonder if we should have the option to use --recreate or not. The additional time to run the tests when I already have all the required dependencies in place may be annoying. I think we should we have the two options: and this can either be 1) make test receiving and passing over the --recreate option (if present) OR make test-no-recreate and make test . I would say the build system should always recreate when validating a change.

scotte:
Unfortunately, without this you aren't testing what you think you are testing - it's not just 3rd party dependencies, but even pyspiffe itself can be out of date in the tox venv, as I discovered.

I think `make test` should always work in a predictable way. If you want to save a few seconds, you can run tox directly via `pipenv run tox -e py36` or so without the `--recreate` option. For local dev, running against all 4 versions of python takes much longer than  the time added by `--recreate`, and for local dev there's no real reason to test all 4 python versions anyway.

glaucimar:
that is true... and I have being using `pipenv run tox -e py36` for my tests so... I am fine with it.

glaucimar (approved):
Thanks, @scotte!

Makefile
```diff
@@ -34,7 +34,7 @@ dev: env
 ## Runs unit tests.
 test: lint
 	@echo "Running unit tests."
-	$(PIPENV_CMD) run tox
+	$(PIPENV_CMD) run tox --recreate
```
Andres-GC, line 37: I think this is not necessary anymore, [PR-40](https://github.com/HewlettPackard/py-spiffe/pull/40) already added the flag and is already merge. (Title does not mention this change but is listed on the description, sorry to mix things on the same PR) 
Is my point correct or I'm missing something?

scotte: Indeed - that change was not in `master` at the time I made this pull request, and it snuck in via unrelated changes. :-) I'll abandon this one. Thanks!

Andres-GC: Yes, it would have been better to add the change in a different PR. Sorry for the confusion. 
Thanks Scott

PR #45, opened 2021-02-16
Fixes #44. Install dependencies into tox venv

### LINKED ISSUE #44: The tox venv doesn't have dependencies from the Pipfile
I found an issue where the tox venv doesn't have all the dependencies specified in the Pipfile. Tox doesn't support Pipfile, but it does support requirements.txt and I found a way to have that generated as part of `make dev` to refresh dependencies. I don't know if there's a better way, but I'll toss out my Pull Request as one approach.

scotte:
I found that the tox venv doesn't have necessary dependencies specified in the Pipfile. Here's one solution, I'm not sure that it's ideal, but it seems to work.

tox doesn't support Pipfile, so this generates a
requirements.txt whenever "make dev" is run based
on the contents of the Pipfile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,8 @@ rm_env:
 dev: env
 	$(PIPENV_CMD) sync --dev && \
 	$(PIPENV_CMD) run pre-commit install
+	mkdir -p .tox
+	$(PIPENV_CMD) lock --requirements >.tox/requirements.txt
 
 
 ## Runs unit tests.
--- a/tox.ini
+++ b/tox.ini
@@ -4,6 +4,7 @@ isolated_build = True
 
 [testenv]
 deps = pytest
+       -r{toxinidir}/.tox/requirements.txt
 commands = python -m pytest test --doctest-modules src
 
 [flake8]


glaucimar:
The way I understand the tox dependencies (as Andres explained to me.. maybe I missed something though) is that the tox will install the runtime packages required by the lab as specified in setup.py, install_requires.

So, if we add a library dependency, we must update the setup.py, "install_requires" to include that dependency. And, of course, after this, recreate the tox environment. I agree having to add this in 2 places (for pip and tox) is annoying but in any case, the change in this PR does not seem to address the need to have the libraries requirements updated in the setup.py file. And if it is updated there, the issue does not "appear".

scotte:
I was unaware of dependencies being in `setup.py` - adding them there might have solved my issue. Regardless, the dependencies in `setup.py` should be automatically generated from the `Pipfile`. Having to add dependencies in two places greatly increases the chance of error, and just as bad, the dependencies in `setup.py` do not contain version information, so what is used for testing and packaging are not guaranteed to be the same, I think.

glaucimar:
Yeap, I agree that putting dependencies in two places is far from ideal so, maybe this PR can update the setup.py. And good question on the versions.

scotte:
I'm going to abandon this and create a new pull request with a better solution that auto-updates `setup.py` - I found an example of how to do so.

PR #47, opened 2021-02-18
Fixes #44. Update setup.py dependencies from Pipfile.lock

### LINKED ISSUE #44: The tox venv doesn't have dependencies from the Pipfile

scotte:
This automatically updates the dependencies in setup.py
at build time using the dependencies from Pipfile.lock.

Signed-off-by: Scott Emmons <scotte@hpe.com>

This works for both `tox` as well as the wheel package, where `requires.txt` looks to contain all the dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,26 @@
 from setuptools import setup, find_packages
 from codecs import open
-from os import path
+from os import path, environ
+from json import load
 
 DESCRIPTION = 'Python library for SPIFFE'
+
+# https://stackoverflow.com/questions/49496994/how-do-i-sync-values-in-setup-py-install-requires-with-pipfile-packages
+def locked_requirements(section):
+    # Explicitly adding an absolute path to PWD looks odd, but there's some
+    # directory shenangians that happen without it
+    with open(path.join(environ['PWD'], 'Pipfile.lock')) as pip_file:
+        pipfile_json = load(pip_file)
+
+    if section not in pipfile_json:
+        print("{0} section missing from Pipfile.lock".format(section))
+        return []
+
+    return [
+        package + detail.get('version', "")
+        for package, detail in pipfile_json[section].items()
+    ]
+
 
 here = path.abspath(path.dirname(__file__))
 
@@ -25,6 +43,6 @@ setup(
     license='Apache License Version 2.0',
     packages=find_packages(where='src', exclude=['test']),
     package_dir={'': 'src'},
-    install_requires=['rfc3987', 'pyjwt', 'cryptography', 'grpcio-tools'],
+    install_requires=locked_requirements('default'),
     python_requires='>=3.6',
 )


glaucimar (approved):
This is great! Thanks, @scotte !

setup.py
```diff
@@ -1,9 +1,27 @@
...
 
+# https://stackoverflow.com/questions/49496994/how-do-i-sync-values-in-setup-py-install-requires-with-pipfile-packages
+def locked_requirements(section):
+    # Explicitly adding an absolute path to PWD looks odd, but there's some
+    # directory shenangians that happen without it
+    with open(path.join(environ['PWD'], 'Pipfile.lock')) as pip_file:
+        pipfile_json = load(pip_file)
+
+    if section not in pipfile_json:
+        print("{0} section missing from Pipfile.lock".format(section))
```
Andres-GC, line 16: Can you update this line to use single quotes please?

setup.py
```diff
@@ -1,9 +1,27 @@
...
+        return []
+
+    return [
+        package + detail.get('version', "")
```
Andres-GC, line 20: I'm wondering if we really want to have an empty version as default value here. These are the libraries that will be required at installation time, so I would say that at least we should force a version we tested.
Any other thoughts about this?

setup.py
scotte: I can - that whole thing is copy/pasted from the link in the source.

setup.py
scotte: Right - so setup.py *does* support entries without a version, in which case you get the latest version. However, this code will use whatever version is in the Pipfile.lock.

I actually agree with you that uncontrolled versions are bad (in fact, that's the whole reason for this PR - to fix the uncontrolled versions we had in setup.py before, which was NOT the same version as in the Pipfile.

It would be nice to make this fail if no version information is present - I'll do so (i.e. latest won't be default), but if I understand what you say above, it won't have any effect because the Pipfile.lock is always going to resolve to the latest version?

I'll modify the quotes, but again that was just copy/paste from a link.

Andres-GC: Yes, Pipfile.lock will resolve to the latest version when there is no version information on the Pipfile. So as you said, the change won't have any affect.

One thing that I realized when reading your response is that I didn't get the whole idea behind this PR.
setup.py does not include version information because it is not consider a best practice, as states in the [packaging docs](https://packaging.python.org/discussions/install-requires-vs-requirements/#install-requires):

```It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.```

What is consider best practice is to indicate any known lower or upper bounds.

Besides version numbers, the `default` section on Pipfile.lock seems to include sub-dependencies too.

So I think I prefer to maintain the `install_requires` manually as it is today. The list of dependencies should be stabilized once we complete the library and shouldn't change much.

I would like to know what others thinks about this and I'm happy to go with whatever the majority decides.

scotte: The problem as I see it is that otherwise we are not testing with same versions of dependencies that will end up getting installed.

I would say if we don't want to take this approach then instead maybe the Pipfile should be created from setup.py, but in either case we probably should not include any versioning information so that the latest version is always used.

I don't see how maintaining separate dependencies chains in both Pipfile and setup.py is a good idea.

Issue #12, opened 2020-12-17
Implement x509_Bundle.Set type


Andres-GC:
`x509_bundle.Set` provides a set of X.509 bundles. It also implements the x509bundle.Source interface.

Issue #8, opened 2020-12-17
Implement x509_bundle.Bundle type


Andres-GC:
The x509_bundle.Bundle type contains trusted X.509 roots for a given trust domain. The bundle marshals/unmarshals from one or more PEM certificate blocks. 
The type needs to be implemented on [x509_bundle.py](https://github.com/HewlettPackard/py-spiffe/blob/master/src/pyspiffe/bundle/x509_bundle/x509_bundle.py):

PR #48, opened 2021-02-19
Implement X509Bundle and X509BundleSet types

maxlambrecht:
Solves #8 and #12 


src/pyspiffe/bundle/x509_bundle/x509_bundle_set.py
```diff
@@ -2,41 +2,61 @@
...
+__all__ = ['X509Bundle']
+
 
 class X509BundleSet(object):
     """X509BundleSet is a set of X509Bundles objects, keyed by trust domain."""
 
-    def __init__(self, bundles: Mapping[TrustDomain, X509Bundle]) -> None:
-        """Creates a new initialized with the given bundles.
+    def __init__(self, bundles_map: Mapping[TrustDomain, X509Bundle]) -> None:
+        """Creates a new initialized Set with the given X509Bundle objects keyed by TrustDomain.
```
Andres-GC, line 17: Would it be more accurate to use `X509BundleSet` instead of only `Set`?

src/pyspiffe/bundle/x509_bundle/x509_bundle_set.py
```diff
@@ -2,41 +2,61 @@
...
+            bundles_map: A map object of X509Bundle objects keyed by TrustDomain to initialize the X509BundleSet.
         """
-        self._bundles = bundles
 
-    def add(self, bundle: X509Bundle):
-        """Adds a new bundle into the set.
+        # create and initialize a dict, so it can be updated (as Mapping doesn't define __setitem__)
+        bundles = {}
+        for pair in bundles_map.items():
+            bundles[pair[0]] = pair[1]
```
Andres-GC, line 26: What do you think about replacing this code by the example below to make it clearer?
```
for td, bundle in bundles_map.items():
    bundles[td] = bundle
```
For each loop iteration, Python will automatically assign the first variable as the key and the second variable as the value for that key.

scotte (dismissed):

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -1,9 +1,261 @@
...
+            x509_cert = x509.load_pem_x509_certificate(
+                cert.as_bytes(), default_backend()
+            )
+            result.add(x509_cert)
+        except Exception:
+            raise ParseX509BundleError('Unable to load PEM X.509 certificate')
+
+    if len(result) < 1:
+        raise ParseX509BundleError('Unable to load PEM X.509 certificate')
+    return result
```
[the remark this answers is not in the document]
maxlambrecht: The semantics are different, the parse certificates method creates a List that represents a chain of Certificates, i.e. a List that is ordered, whereas in the case of the authorities the order doesn't matter and what matters is only that each authority is not duplicated in the Set.

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -1,9 +1,261 @@
...
+            encoding: Bundle encoding format, either serialization.Encoding.PEM or serialization.Encoding.DER
+
+        Raises:
+            ValueError: In case the encoding is not either PEM or DER (from serialization.Encoding)
+            X509BundleError: In case the authorities in the bundle cannot be converted to bytes.
+            SaveX509BundleError: In the case the file path in bundle_path cannot be open to write, or there is an error
+                                writing the authorities bytes to the file.
+        """
+
+        if not (encoding is encoding.PEM or encoding is encoding.DER):
```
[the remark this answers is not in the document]
maxlambrecht: Yes.

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -1,9 +1,261 @@
...
+        raise LoadX509BundleError(
+            'Certs chain file file not found: {}'.format(certs_chain_path)
+        )
+    except Exception as err:
+        raise LoadX509BundleError(
+            'Certs chain file could not be read: {}'.format(str(err))
+        )
+
+
+def _write_certs_to_file(
```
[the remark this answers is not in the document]
maxlambrecht: In this case I probably agree, and we could refactor this once both PRs are merged, moving the "duplicate" functionality to some common module.

src/pyspiffe/bundle/x509_bundle/x509_bundle_set.py
maxlambrecht: that's indeed a good improvement to my newbie coding style :)

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -1,9 +1,261 @@
...
+    def x509_authorities(self) -> Set[Certificate]:
+        """Returns the set of X.509 authorities in the bundle. """
+        return self._x509_authorities
+
+    def add_authority(self, x509_authority: Certificate) -> None:
+        """Adds an X.509 authority to the bundle. """
+        if not self._x509_authorities:
+            self._x509_authorities = set()
+
+        self._x509_authorities.add(x509_authority)
```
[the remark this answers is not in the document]
maxlambrecht: Makes sense.

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -1,9 +1,261 @@
...
+from pyasn1.codec.der.decoder import decode
+
+_BUNDLE_FILE_MODE = 0o644
+
+__all__ = ['X509Bundle']
+
 
 class X509Bundle(object):
-    """X509Bundle is a collection of trusted public key material for a trust domain."""
+    """
```
glaucimar, line 30: nit: text should be in the same line as """

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -1,9 +1,261 @@
...
+
+    def add_authority(self, x509_authority: Certificate) -> None:
+        """Adds an X.509 authority to the bundle. """
+        self._x509_authorities.add(x509_authority)
+
+    def remove_authority(self, x509_authority: Certificate) -> None:
+        """Removes an X.509 authority from the bundle. """
+        if not self._x509_authorities:
+            return
+        self._x509_authorities.remove(x509_authority)
```
glaucimar, line 66: nit: If the authority is not present in the Set, it will raise KeyError , I suggest this gets added to the method docString.

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -1,9 +1,261 @@
...
+
+        Returns:
+            An instance of 'X509Bundle' with the X.509 authorities associated to the given trust domain.
+
+        Raises:
+            ParseBundleError: In case the set of x509_authorities cannot be parsed from the bundle_bytes.
+        """
+
+        if not trust_domain:
+            raise X509BundleError(EMPTY_DOMAIN_ERROR)
```
glaucimar, line 84: nit: add X509BundleError to the docstring? A cpuple of other ocurrences of the same.

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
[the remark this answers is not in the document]
Andres-GC: Ok, thanks for the explanation. 
I still think there is place for a refactor and we can unify that code in one function that returns an ordered list. Then for the case where you need to remove duplicates we can use `s = set(list)`.
As you mentioned in another comment, let's see this again once both PRs are merged. Later on, we can work together in a refactor if necessary.
Thanks Max

Andres-GC (dismissed):

test/bundle/x509bundle/test_x509_bundle.py
```diff
@@ -0,0 +1,254 @@
...
+    bundle.add_authority(x509_cert_1)
+    bundle.add_authority(x509_cert_2)
+
+    assert len(bundle.x509_authorities()) == 2
+
+    for a in bundle.x509_authorities():
+        assert isinstance(a, Certificate)
+        assert 'CN=PEMUTILTEST' in a.subject.rfc4514_string()
+
+    bundle.remove_authority(x509_cert_1)
```
glaucimar, line 245: nit: what about removing an unexistent Certificate and adding one that is already there?

glaucimar (commented):
Some small points to be looked at.

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
maxlambrecht: It doesn't raise an error, I changed the `test_add_remove_authority` to prove that.

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
maxlambrecht: Added the error to the docstrings and also moved the validation to the `__init__` method as it was repeated in multiple places and notably missing in the constructor.

test/bundle/x509bundle/test_x509_bundle.py
maxlambrecht: Added it to the test: add and remove multiple times the same certificate.

glaucimar (approved):
Thanks for addressing the issues, Max!

Andres-GC (approved):

PR #51, opened 2021-02-25
Update CONTRIBUTING.md

glaucimar:
This PR makes small improvements to CONTRIBUTING.md file to:
1- make the text about Attributes more specific
2- change the class attributes names in the example to a name that is different from the instance attributes names.
3- include the declaration of the class attributes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,7 +108,7 @@ SPIFFE_ID_MAXIMUM_LENGTH = 2048
 ```
 
 ### Classes
-Class docstrings should document the list of public members including its types.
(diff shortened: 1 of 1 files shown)


maxlambrecht (commented):
Can we add here the convention we are following about underscoring the "meant to be private" methods and attributes ? Or maybe it could be in separate PR.

glaucimar:
> Can we add here the convention we are following about underscoring the "meant to be private" methods and attributes ? Or maybe it could be in separate PR.

Done!

maxlambrecht (approved):
LGTM

scotte (approved):

PR #57, opened 2021-03-05
Adds lint and test coverage report jobs to CI

Andres-GC:
This PR:
- Adds the test coverage of the project to the CI and post the results to [coveralls.io](https://coveralls.io/github/HewlettPackard/py-spiffe).
- Adds make task `test_cov` to generate test coverage and it won't try to post the results to the coveralls site.
- Adds `lint` to the CI. When running on the CI it will be executed using a `tox virtual environment` but when it is run locally it continues outside of `tox` because is faster. 
- Configures Coverage Threshold for failure at [coveralls.io](https://coveralls.io/github/HewlettPackard/py-spiffe) site. Test coverage now is 91.4%. New builds will fail if test coverage is less than 90%.

Note: In order to post coverage reports to `coveralls.io`, we need to set the Env Var `COVERALLS_REPO_TOKEN` on CircleCi (is set as project level). CircleCi by default won't share Env Var with builds generated by PRs coming from forked repos. To solved this issue, the option `Project Settings -> Advance -> Pass secrets to builds from forked pull requests` is set to ON now. This is the only configuration we have so far. If in the future we have any other ENV VAR set, we should review this config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,18 +10,31 @@ commands:
       env:
         default: "py36"
         type: string
+      coveralls_token:
+        default: ""
+        type: string
     steps:
-      - run: tox --recreate -e <<parameters.env>>
-
+      - run:
+          name: run tox
+          command: |
+            COVERALLS_REPO_TOKEN=<<parameters.coveralls_token>> tox --recreate -e <<parameters.env>>
 jobs:
-  test_py36:
+  lint:
     docker:
       - image: circleci/python:3.6
     steps:
       - checkout
       - install_tox
       - run_tox:
-          env: "py36"
+          env: "lint"
+
+  test_py36:
+    docker:
+      - image: circleci/python:3.6
+    steps:
+      - checkout
+      - install_tox
+      - run_tox
 
   test_py37:
     docker:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -10,18 +10,33 @@ commands:
...
+        type: string
     steps:
-      - run: tox --recreate -e <<parameters.env>>
-
+      - run:
+          name: run tox
+          command: |
+            echo "export COVERALLS_REPO_TOKEN=<<parameters.coveralls_token>>" >> $BASH_ENV
+            source $BASH_ENV
+            tox --recreate -e <<parameters.env>>
```
scotte, line 22: An easier and more recommended way that doesn't affect the environment for subsequent commands is to do the following, assuming circleci just passes the command to the shell directly:

`COVERALLS_REPO_TOKEN=<<parameters.coveralls_token>> tox --recreate -e <<parameters.env>>`

Obviously this won't work if the token is used later, in which case ignore this comment. :-)

scotte (approved):
Looks good - one comment which you can ignore if you like or wouldn't work anyway.

.circleci/config.yml
Andres-GC: You are right. That is much clearer and exactly what I need here. The env var is only used to run Tox so it is perfect.
I'm pretty sure I tried a similar approach using directly the env var and it didn't work, the env var was not set to the context when calling tox. That is why I ended up using the parameter here. 
I did not think about using the parameter directly as you suggested. I'm glad you noticed that because it looks much better now.
 
Thanks Scott.

glaucimar (approved):

scotte (approved):

glaucimar (approved):
thanks Andres!

PR #59, opened 2021-03-17
Enforce immutability and thread-safety on X.509 SVIDs and Bundle objects

maxlambrecht:
This PR makes the types `X509Svid`, `X509Bundle`, and `X509BundleSet` immutable and thread-safe by making the getters returns copies of the internal collections, making the constructors deepcopy the collections passed as arguments, and making the `add` and `remove` methods use a Lock to prevent data-races.

Once this PR gets the approval of the team, there will remain to make the same kind of adaptations to the other types (`JWTBundle`, etc)

### DIFF
--- a/src/pyspiffe/bundle/x509_bundle/x509_bundle.py
+++ b/src/pyspiffe/bundle/x509_bundle/x509_bundle.py
@@ -57,9 +57,9 @@ class X509Bundle(object):
             self._x509_authorities = set()
 
     def __eq__(self, o: object) -> bool:
+        if not isinstance(o, X509Bundle):
+            return False
         with self.lock:
-            if not isinstance(o, X509Bundle):
-                return False
             return (
                 self._trust_domain.__eq__(o._trust_domain)
                 and self._x509_authorities == o._x509_authorities
@@ -76,7 +76,8 @@ class X509Bundle(object):
 
     def add_authority(self, x509_authority: Certificate) -> None:
         """Adds an X.509 authority to the bundle. """
-        self._x509_authorities.add(copy.copy(x509_authority))
+        with self.lock:
+            self._x509_authorities.add(copy.copy(x509_authority))
 
     def remove_authority(self, x509_authority: Certificate) -> None:
         """Removes an X.509 authority from the bundle. """
--- a/src/pyspiffe/bundle/x509_bundle/x509_bundle.py
+++ b/src/pyspiffe/bundle/x509_bundle/x509_bundle.py
@@ -51,7 +51,7 @@ class X509Bundle(object):
             raise X509BundleError(EMPTY_DOMAIN_ERROR)
         self._trust_domain = trust_domain
 
-        if x509_authorities is not None:
+        if x509_authorities:
             self._x509_authorities = x509_authorities.copy()
         else:
             self._x509_authorities = set()
@@ -99,7 +99,7 @@ class X509Bundle(object):
 
         Raises:
             X509BundleError: In case the trust_domain is empty.
-            ParseBundleError: In case the set of x509_authorities cannot be parsed from the bundle_bytes.
+            ParseX509BundleError: In case the set of x509_authorities cannot be parsed from the bundle_bytes.
         """
 
         try:
--- a/src/pyspiffe/bundle/x509_bundle/x509_bundle.py
+++ b/src/pyspiffe/bundle/x509_bundle/x509_bundle.py
@@ -1,7 +1,6 @@
 """
 This module manages X.509 Bundle objects.
 """
-import copy
 import os
 import threading
 from typing import Set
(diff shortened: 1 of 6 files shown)


Andres-GC (approved):
Thanks Max!

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -44,31 +45,46 @@ def __init__(
...
-        if not self._x509_authorities:
-            return
-        self._x509_authorities.remove(x509_authority)
+        with self.lock:
+            if not self._x509_authorities:
+                return
+            self._x509_authorities.remove(x509_authority)
 
     @classmethod
     def parse(cls, trust_domain: TrustDomain, bundle_bytes: bytes) -> 'X509Bundle':
```
glaucimar, line 90: I understand this is not a change in this PR but this method raises ParseX509BundleError as well and this is not in the docstring.

src/pyspiffe/bundle/x509_bundle/x509_bundle_set.py
```diff
@@ -13,26 +14,24 @@
...
 
     def put(self, bundle: X509Bundle) -> None:
         """Adds a new X509Bundle object or replace an existing one into the set.
 
         Args:
             bundle: The new X509Bundle to put into the set.
         """
-        self._bundles[bundle.trust_domain()] = bundle
+        with self.lock:
+            self._bundles[bundle.trust_domain()] = bundle
```
glaucimar, line 34: should this method store a copy of the input bundle object?

src/pyspiffe/bundle/x509_bundle/x509_bundle_set.py
```diff
@@ -46,7 +45,8 @@ def get_x509_bundle_for_trust_domain(
             A X509Bundle object for the given TrustDomain.
             None if the TrustDomain is not found in the set.
         """
-        return self._bundles.get(trust_domain)
+        with self.lock:
+            return self._bundles.get(trust_domain)
```
glaucimar, line 49: should this return a copy?

glaucimar (commented):
Thanks for this, @maxlambrecht ! I have added some questions and considerations. Please let me know.

glaucimar (dismissed):
Thanks @maxlambrecht, it looks great!

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -44,31 +45,46 @@ def __init__(
...
         """
+
+        self.lock = threading.Lock()
+
         if not trust_domain:
             raise X509BundleError(EMPTY_DOMAIN_ERROR)
         self._trust_domain = trust_domain
 
-        self._x509_authorities = x509_authorities
+        self._x509_authorities = copy.deepcopy(x509_authorities)
```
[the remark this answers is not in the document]
maxlambrecht: Thanks for the suggestion!

src/pyspiffe/bundle/x509_bundle/x509_bundle_set.py
maxlambrecht: As the bundle itself "protects" its internal collections through safe access methods, I think it is safe to share bundles and this way we avoid incurring in overheads needing to deep copy everything in each put/get method.

src/pyspiffe/bundle/x509_bundle/x509_bundle_set.py
maxlambrecht: Same logic as previous comment.

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
maxlambrecht: The name of the thrown exception is wrong in the docstring.

src/pyspiffe/bundle/x509_bundle/x509_bundle_set.py
```diff
@@ -13,26 +13,28 @@
...
-        bundles = {}
-        for td, bundle in bundles_map.items():
-            bundles[td] = bundle
-        self._bundles = bundles
+        self.lock = threading.Lock()
+
+        if bundles_map:
+            self._bundles = bundles_map.copy()
+        else:
+            self._bundles = dict()
```
Andres-GC, line 28: nit: I think we should rename the parameter to `bundle_dict`.

Andres-GC (approved):

src/pyspiffe/bundle/x509_bundle/x509_bundle_set.py
maxlambrecht: or maybe it could be just `bundles`, because why repeat the type in the name when it's already explicit, right?

glaucimar (approved):

PR #64, opened 2021-03-26
Intial Implementation of JWTBundleSet

Andres-GC:
Adds an initial implementation of JWTBundleSet
Makes JwtBundle operation thread-safe
Fixes some typos and a variable name.

### DIFF
--- a/src/pyspiffe/bundle/jwt_bundle/exceptions.py
+++ b/src/pyspiffe/bundle/jwt_bundle/exceptions.py
@@ -4,22 +4,34 @@ This module handles JWT bundle exceptions.
 from pyspiffe.exceptions import PySpiffeError
 
 
-class AuthorityNotFoundError(PySpiffeError):
-    """Raised when an authority is not found associated with a key_id.
+class JwtBundleError(PySpiffeError):
+    """Top level exception for the JwtBundle module. """
 
-    Attributes:
-        message: Message describing the error.
-    """
+    def __init__(self, message: str) -> None:
+        """Creates an instance of JwtBundleError.
+
+        Args:
+            message: Message describing the error.
+        """
+
+        self.message = message
+
+    def __str__(self) -> str:
+        return self.message
+
+
+class AuthorityNotFoundError(JwtBundleError):
+    """Raised when an authority is not found associated with a key_id."""
 
     _MESSAGE = 'Key ({}) not found in authorities.'
 
-    def __init__(self, key_id: str = "not specified") -> None:
+    def __init__(self, key_id: str = 'not specified') -> None:
         """Creates an instance of AuthorityNotFoundError.
 
         Args:
-            key_id: the key_id with no authority associated.
+            key_id: The key_id with no authority associated.
         """
-        self.message = self._MESSAGE.format(key_id)
+        super().__init__(self._MESSAGE.format(key_id))
 
     def __str__(self) -> str:
         return self.message
--- a/src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
+++ b/src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
@@ -1,10 +1,12 @@
 """
 JwtBundle module manages JwtBundle objects.
 """
-from pyspiffe.bundle.jwt_bundle.exceptions import AuthorityNotFoundError
-from pyspiffe.spiffe_id.trust_domain import TrustDomain
-from typing import Dict, Union
+import threading
+from typing import Dict, Union, Optional
 from cryptography.hazmat.primitives.asymmetric import ec, rsa, dsa, ed25519, ed448
+
+from pyspiffe.spiffe_id.trust_domain import TrustDomain, EMPTY_DOMAIN_ERROR
+from pyspiffe.bundle.jwt_bundle.exceptions import JwtBundleError
 
 _PUBLIC_KEY_TYPES = Union[
     dsa.DSAPublicKey,
@@ -28,15 +30,26 @@ class JwtBundle(object):
 
         Args:
             trust_domain: The TrustDomain to associate with the JwtBundle instance.
-            jwt_authorities: A dictionay with key_id->PublicKey valid for the given TrustDomain.
+            jwt_authorities: A dictionary with key_id->PublicKey valid for the given TrustDomain.
         """
-        self._trust_domain = trust_domain
-        if jwt_authorities:
-            self._jwt_authorities = jwt_authorities.copy()
-        else:
-            self._jwt_authorities = {}
+        self.lock = threading.Lock()
 
-    def find_jwt_authority(self, key_id: str) -> _PUBLIC_KEY_TYPES:
+        if not trust_domain:
+            raise JwtBundleError(EMPTY_DOMAIN_ERROR)
+
+        self._trust_domain = trust_domain
+        self._jwt_authorities = jwt_authorities.copy() if jwt_authorities else {}
+
+    def trust_domain(self) -> TrustDomain:
+        """Returns the trust domain of the bundle. """
+        return self._trust_domain
+
+    def jwt_authorities(self) -> Dict[str, _PUBLIC_KEY_TYPES]:
+        """Returns a copy of JWT authorities in the bundle."""
+        with self.lock:
+            return self._jwt_authorities.copy()
+
+    def get_jwt_authority(self, key_id: str) -> Optional[_PUBLIC_KEY_TYPES]:
         """Returns the authority for the specified key_id.
 
         Args:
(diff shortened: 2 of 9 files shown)


src/pyspiffe/bundle/jwt_bundle/exceptions.py
```diff
@@ -19,7 +31,7 @@ def __init__(self, key_id: str = "not specified") -> None:
         Args:
             key_id: the key_id with no authority associated.
```
glaucimar, line 32: nit: I know this is not your change but I found this and so... kei_id: The key_id ..... it should be capital T.

src/pyspiffe/bundle/jwt_bundle/exceptions.py
```diff
@@ -4,12 +4,24 @@
...
+        self.message = message
+
+    def __str__(self) -> str:
+        return self.message
+
+
+class AuthorityNotFoundError(JwtBundleError):
+    """Raised when an authority is not found associated with a key_id."""
 
     _MESSAGE = 'Key ({}) not found in authorities.'
```
glaucimar, line 27: nit: all code is using single quotes, we should use here too. by here, I mean the default parameter value for str; str = 'not specified'.

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
```diff
@@ -51,7 +65,10 @@ def find_jwt_authority(self, key_id: str) -> _PUBLIC_KEY_TYPES:
...
             raise ValueError('key_id cannot be empty.')
 
-        key = self._jwt_authorities.get(key_id, None)
-        if not key:
-            raise AuthorityNotFoundError(key_id)
-        return key
+        with self.lock:
+            key = self._jwt_authorities.get(key_id)
+
+            if not key:
```
glaucimar, line 71: Is it really need to keep the lock until here?

glaucimar (dismissed):
Great job, @Andres-GC !

maxlambrecht (commented):
Thanks Andres for this, I left some comments.

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
```diff
@@ -28,13 +31,24 @@ def __init__(
...
+
+        if not trust_domain:
+            raise JwtBundleError(EMPTY_DOMAIN_ERROR)
+
         self._trust_domain = trust_domain
-        if jwt_authorities:
-            self._jwt_authorities = jwt_authorities.copy()
-        else:
-            self._jwt_authorities = {}
+        self._jwt_authorities = jwt_authorities.copy() if jwt_authorities else {}
```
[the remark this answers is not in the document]
glaucimar: yeah... I liked that too ... so, wanted to come here and second Max :)

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
```diff
@@ -51,7 +65,10 @@ def find_jwt_authority(self, key_id: str) -> _PUBLIC_KEY_TYPES:
...
 
-        key = self._jwt_authorities.get(key_id, None)
-        if not key:
-            raise AuthorityNotFoundError(key_id)
-        return key
+        with self.lock:
+            key = self._jwt_authorities.get(key_id)
+
+            if not key:
+                raise AuthorityNotFoundError(key_id)
```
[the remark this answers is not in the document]
glaucimar: this is indeed a good point Max... in either way, I think we need to be consistent in get_jwt_bundle_for_trust_domain and here... (unless, of course, there is a good reason not to). Deserves some research and I don't know the answer to the most pythonic way :).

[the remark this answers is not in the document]
Andres-GC: More than if it's pythonic or not pythonic, I think the point here is where we want to handle an empty response.
Not only `get_jwt_bundle_for_trust_domain` handles this scenario differently, `x509Bundle.get_x509_bundle_for_trust_domain()`  also do so.
I'll make this change so we are consistent. Thanks.

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
Andres-GC: This will be fixed as part of the changes from the other comment (raising and error vs making the return optional).
But I would like to know what do you guys think about renaming this method from `find_jwt_authority()` to `get_jwt_authority`?
For me, `get` is more accurate in this scenario where `_jwt_authority` is a dictionary.

maxlambrecht: I like `get_`

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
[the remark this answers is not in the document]
maxlambrecht: Sure!

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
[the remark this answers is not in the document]
maxlambrecht: I think throwing an Exception in case of missing a key_id is something that we brought here from Java, as in the java world returning an Optional is still not generally accepted, but I feel comfortable using it here.

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
glaucimar: I like get :)... if we have put(), it makes sense to have get()... if we want get_jwt_authority, I would say let's have put_jwt_authority too. :). I prefer put/get.

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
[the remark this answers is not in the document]
glaucimar: Max, agree... the get from dict returns a default value in case key is not present... [key] raises an exception. Seems to me this should be aligned to the get behaviour of Dict. If we proceed down this road, should we also allow for the caller to inform the default to use in case key is not present? I like this capability.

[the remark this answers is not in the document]
maxlambrecht: yeah, the get from the dictionary returns None in case of a missing key, that's the default value, except that you specify another default value. Not sure if we need to provide a `get_jwt_authority(key, default_authority)`, probably not, it's not present in the other libraries.

```
def test_missing_key():
    mydict = {}
    assert mydict.get('mising key') is None
```

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
Andres-GC: Actually `JwtBundleSet is a dictionary of JWTBundles objects, keyed by trust domain`. Why don't we just call the methods `get` & `put`?

`JwtBundleSet.get_jwt_bundle_for_trust_domain()`
vs
`JwtBundleSet.get()`

Each method has its own docstring to make its purpose clear.

src/pyspiffe/bundle/jwt_bundle/jwt_bundle_set.py
```diff
@@ -2,25 +2,26 @@
...
 from pyspiffe.spiffe_id.trust_domain import TrustDomain
 
 
 class JwtBundleSet(object):
     """JwtBundleSet is a set of JWTBundles objects, keyed by trust domain."""
 
-    def __init__(self, bundles: Mapping[TrustDomain, JwtBundle]) -> None:
-        """Creates a new initialized with the given JWT bundles.
+    def __init__(self, bundles: Dict[TrustDomain, JwtBundle]) -> None:
+        """Creates a new initialized with the given JWT bundles objects keyed by TrustDomain.
```
scotte, line 15: Nit: awkward sentence, maybe "Creates a new set initialized with the given JWT bundle objects keyed by TrustDomain"?

scotte (approved):

scotte (approved):

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
```diff
@@ -51,7 +64,5 @@ def find_jwt_authority(self, key_id: str) -> _PUBLIC_KEY_TYPES:
         if not key_id:
             raise ValueError('key_id cannot be empty.')
 
-        key = self._jwt_authorities.get(key_id, None)
-        if not key:
-            raise AuthorityNotFoundError(key_id)
-        return key
+        with self.lock:
+            return self._jwt_authorities.get(key_id)
```
glaucimar, line 68: It does not raise AuthorityNotFoundError anymore, it only raises ValueError. need to fix docstring.

glaucimar (approved):
Thanks, Andres!

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
Andres-GC: Good catch Glaucimar. It is fixed now. I also added a note to indicate that the method returns None if the key_id is not found.

glaucimar (approved):

maxlambrecht (approved):
💯

PR #65, opened 2021-03-28
Fixes on jwt_svid parse and validate.

glaucimar:
- fixing the missing sub claim in token, adidng test for it, adding typing for test function

### DIFF
--- a/src/pyspiffe/svid/jwt_svid.py
+++ b/src/pyspiffe/svid/jwt_svid.py
@@ -114,7 +114,7 @@ class JwtSvid(object):
             token_header = jwt.get_unverified_header(token)
             validator = JwtSvidValidator()
             validator.validate_header(token_header)
-            signing_key = jwt_bundle.find_jwt_authority(token_header.get('kid', None))
+            signing_key = jwt_bundle.find_jwt_authority(token_header.get('kid'))
 
             public_key_pem = signing_key.public_bytes(
                 encoding=serialization.Encoding.PEM,
@@ -123,7 +123,7 @@ class JwtSvid(object):
 
             claims = jwt.decode(
                 token,
-                algorithms=token_header.get('alg', None),
+                algorithms=token_header.get('alg'),
                 key=public_key_pem,
                 audience=audience,
                 options={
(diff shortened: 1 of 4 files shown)


src/pyspiffe/svid/jwt_svid.py
```diff
@@ -132,7 +132,7 @@ def parse_and_validate(
                     'verify_exp': True,
                 },
             )
-            spiffe_ID = SpiffeId.parse(claims['sub'])
+            spiffe_ID = SpiffeId.parse(claims.get('sub', None))
```
Andres-GC, line 135: nit: [`get()`](https://docs.python.org/3/library/stdtypes.html#dict.get) default value is `None` so it is not necessary to add it here.

test/svid/jwtsvid/test_jwt_svid.py
```diff
@@ -231,6 +231,29 @@ def test_parse_and_validate_invalid_missing_kid_header():
     assert str(exception.value) == 'key_id cannot be empty.'
 
 
+def test_parse_and_validate_invalid_missing_sub():
+    rsa_key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
+    jwt_bundle = JwtBundle(TrustDomain('test.orgcom'), {'kid1': rsa_key.public_key()})
+    rsakeypem, _ = get_keys_pems(rsa_key)
```
Andres-GC, line 237: nit: `rsa_key_pem` looks clearer to me than `rsakeypem`.

An alternative to  `rsa_key_pem, _ = get_keys_pems(rsa_key)`is: `rsa_key_pem = get_keys_pems(rsa_key)[1]`

`get_keys_pems()` returns two values what means that is returning a tuple. That is why you can access what you need here using the index inside that returned tuple. I think this is it _pythonic_ way to do it.

Andres-GC (dismissed):

src/pyspiffe/svid/jwt_svid.py
glaucimar: Thanks!

test/svid/jwtsvid/test_jwt_svid.py
glaucimar: I understand this... but I have seem both in Python, so, I am not so sure this is the _pythonic_ way... I will make the variable change name.

Andres-GC (approved):
Thanks Glaucimar

maxlambrecht (approved):

PR #66, opened 2021-03-29
Upgrade grpc dependencies to version 1.36.1

maxlambrecht:
Upgrade to the latest grpc version.


glaucimar (dismissed):

Pipfile
```diff
@@ -18,8 +18,8 @@ pem = "~=21.1.0 "
 python_version = "3.6"
 
 [dev-packages]
-grpcio = "~=1.33.2"
-grpcio-tools = "~=1.33.2"
+grpcio = "~=1.36.1"
+grpcio-tools = "~=1.36.1"
```
Andres-GC, line 22: These lines here are not necessary. The libraries are already listed in the `packages` section what is correct for our library.
Can you delete them as part of this PR please?

Andres-GC (approved):
💯

glaucimar (approved):

PR #77, opened 2021-04-05
Improves test coverage

Andres-GC:
This PRs:
- Improves test coverage.
- Adds MissingClaimError to make clear difference between an invalid claim value and a missing required claim.
- Fixes some typos.

### DIFF
--- a/src/pyspiffe/svid/jwt_svid_validator.py
+++ b/src/pyspiffe/svid/jwt_svid_validator.py
@@ -13,6 +13,7 @@ from pyspiffe.svid.exceptions import (
     InvalidClaimError,
     InvalidAlgorithmError,
     InvalidTypeError,
+    MissingClaimError,
 )
 
 AUDIENCE_NOT_MATCH_ERROR = 'audience does not match expected value'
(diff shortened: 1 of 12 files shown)


src/pyspiffe/svid/jwt_svid_validator.py
```diff
@@ -77,38 +78,36 @@ def validate_claims(
...
             if not payload.get(claim):
-                raise InvalidClaimError(claim)
-        aud = payload.get('aud')
-        if aud is not None:
-            self._validate_exp(str(payload.get('exp')))
-            self._validate_aud(aud, expected_audience)
+                raise MissingClaimError(claim)
+
+        self._validate_exp(str(payload.get('exp')))
+        self._validate_aud(payload.get('aud', []), expected_audience)
```
Andres-GC, line 98: Here is where I changed the exception to make a difference between missing a required claim vs an invalid value.

Besides that, `aud` is part of the `_REQUIRED_CLAIMS` so `if aud is not None:` makes no sense. I removed it. Please, let me know if I'm missing something here.

glaucimar (commented):
thanks for the PR.

src/pyspiffe/svid/jwt_svid_validator.py
```diff
@@ -44,21 +45,21 @@ def __init__(self) -> None:
         pass
 
     def validate_header(self, header: Dict[str, str]) -> None:
```
maxlambrecht, line 47: reading the docstring I think that this method should be renamed to `validate_headers` and the dict param to `headers`. Would you do that as part of this PR?

maxlambrecht (commented):
Great improvements @Andres-GC !

src/pyspiffe/svid/jwt_svid_validator.py
Andres-GC: Sure. I'll update the PR with that change.
Thanks Max.

glaucimar: I don't agree with this change.. The token has 1 header and multiple key-values inside. And here, this method is validating if the header has the required values. The header is provided as input for the method (in the form of Dic) containing the key-value pairs that composes the token's header.

maxlambrecht: The SPIFFE standard refers the parts of the token header as, for example, `alg` header, `typ` header, and as we are validating a couple of them I thought it would make sense to rename it as `validate_headers`. On the other hand, the RFC 7518 only uses `header` as a whole and then talks about header parameters to refer the specific parts. So, either way is ok for me, your call @glaucimar ;)

Andres-GC: `def validate_header(self, parameters: Dict[str, str]) -> None:` would work here? What do you think?

glaucimar: > `def validate_header(self, parameters: Dict[str, str]) -> None:` would work here? What do you think?

Ah.. to me, that makes sense.

glaucimar: > The SPIFFE standard refers the parts of the token header as, for example, `alg` header, `typ` header, and as we are validating a couple of them I thought it would make sense to rename it as `validate_headers`. On the other hand, the RFC 7518 only uses `header` as a whole and then talks about header parameters to refer the specific parts. So, either way is ok for me, your call @glaucimar ;)

Thanks @maxlambrecht for bringing all of this information.

maxlambrecht: Sounds good

Andres-GC: Ok, I made che renamed and also renamed the variable used to pass the parameter to the function based on the docstrings from `jwt.get_unverified_header(token)`:
```
    def get_unverified_header(self, jwt):
        """Returns back the JWT header parameters as a dict()

        Note: The signature is not verified so the header parameters
        should not be fully trusted until signature verification is complete
        """
        headers = self._load(jwt)[2]
        self._validate_headers(headers)

        return headers
```

Also in this new commit I updated some test cases according to a recent change where a parameter (audience) change its type, from List to Set.

glaucimar (approved):
Thanks, Andres!

PR #81, opened 2021-04-12
Implements ValidateJWTSVID(). 

Andres-GC:
This PR also updates the code to use a set instead of a list as parameters when calling `JwtSvid()`, `JwtSvid.parse_insecure()` and `JwtSvid.pase_and_validate()`. This where required based on a previous update.

### DIFF
--- a/src/pyspiffe/workloadapi/default_workload_api_client.py
+++ b/src/pyspiffe/workloadapi/default_workload_api_client.py
@@ -1,7 +1,7 @@
 """
 This module provides a Workload API client.
 """
-from typing import Optional, List, Set, Mapping
+from typing import Optional, List, Mapping
 
 import grpc
 
@@ -20,6 +20,7 @@ from pyspiffe.workloadapi.exceptions import (
     FetchX509SvidError,
     FetchX509BundleError,
     FetchJwtSvidError,
+    ValidateJwtSvidError,
 )
 from pyspiffe.workloadapi.grpc import header_manipulator_client_interceptor
 from pyspiffe.svid.x509_svid import X509Svid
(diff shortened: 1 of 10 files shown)


maxlambrecht (dismissed):
LGTM!

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -184,8 +184,17 @@ def validate_jwt_svid(self, token: str, audience: str) -> JwtSvid:
         Returns:
             JwtSvid: If the token and audience could be validated.
         """
+        try:
```
glaucimar, line 187: why not test the token and audience provided? at least if they are not None or empty before sending these to the agent? I know these validations are performed at the agent side but we shall be able to provide a quicker error if we avoid the agent call.

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -184,8 +184,17 @@ def validate_jwt_svid(self, token: str, audience: str) -> JwtSvid:
...
+                audience=audience,
+                svid=token,
+            )
 
-        pass
+            self._spiffe_workload_api_stub.ValidateJWTSVID(request)
+        except Exception as e:
+            raise ValidateJwtSvidError(str(e))
+
+        return JwtSvid.parse_insecure(token, {audience})
```
glaucimar, line 197: parse_insecure may raise a bunch of exceptions, I understand this is highly not likely to be thrown after the agent validation BUT I am wondering if we should either have these identified in the docstring for this method OR treat the highly unlikely case these are raised.

test/workloadapi/test_default_workload_api_client_jwt.py
```diff
@@ -115,36 +112,59 @@ def test_fetch_jwt_svid_wrong_token(mocker):
...
+    assert svid.spiffe_id == SpiffeId.parse(spiffe_id)
+    assert svid.token == jwt_svid
+    assert svid.claims['aud'] == [audience]
+    assert svid.audience == {audience}
+
+
+def test_validate_jwt_svid_raise_error(mocker):
+    audience = 'spire'
+    spiffe_id = 'spiffe://test.com/my_service'
+    jwt_svid = create_jwt(rsa_key_pem, 'kid1', 'RS256', [audience], spiffe_id)
```
glaucimar, line 161: in this case, you do not need any of these parameters as there is no validation in the method itself.....

glaucimar (commented):
Thanks Andres!

src/pyspiffe/workloadapi/default_workload_api_client.py
Andres-GC: `parse_insecure()` raises exception when the token is invalid. If that is the case, the workloadApi will return an error response and `parse_insecure()` will never be called. I think there is no need to update the docstrings here.

maxlambrecht (approved):
💯

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -184,8 +184,22 @@ def validate_jwt_svid(self, token: str, audience: str) -> JwtSvid:
         Returns:
             JwtSvid: If the token and audience could be validated.
         """
```
glaucimar, line 186: Missing Raises statement in the docstring.

glaucimar (approved):
Thanks Andres! I missed the docstring info for the raised errors. Feel free to add that change to another PR.

src/pyspiffe/workloadapi/default_workload_api_client.py
Andres-GC: Good Catch! I'll merge this PR and add the docstring in the other open PR that has to be updated.
Thanks.

PR #82, opened 2021-04-13
All exception messages end up with a dot automatically.

Andres-GC:
This PR updates the errors thrown by the library to include a dot at the end of the message automatically.

The changes:
* All custom exceptions inherit from `PySpiffeError`, the top level exception for the library.
* The final dot in the message is not necessary when raising an exception.
* To be consistent in the code, the new exception `ArgumentError` is created to replace all occurrences of `ValueError`.

This way, all error messages in the code looks the same and there is no need for a final dot. I removed the final dot from all places to make things clearer.

### DIFF
--- a/src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
+++ b/src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
@@ -7,6 +7,7 @@ from cryptography.hazmat.primitives.asymmetric import ec, rsa, dsa, ed25519, ed4
 
 from pyspiffe.spiffe_id.trust_domain import TrustDomain, EMPTY_DOMAIN_ERROR
 from pyspiffe.bundle.jwt_bundle.exceptions import JwtBundleError
+from pyspiffe.exceptions import ArgumentError
 
 _PUBLIC_KEY_TYPES = Union[
     dsa.DSAPublicKey,
@@ -60,10 +61,10 @@ class JwtBundle(object):
             None if the key_id is not found.
 
         Raises:
-            ValueError: When key_id is not valid (empty or None).
+            ArgumentError: When key_id is not valid (empty or None).
         """
         if not key_id:
-            raise ValueError('key_id cannot be empty.')
+            raise ArgumentError('key_id cannot be empty')
 
         with self.lock:
             return self._jwt_authorities.get(key_id)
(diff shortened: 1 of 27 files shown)


src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
```diff
@@ -60,10 +61,10 @@ def get_jwt_authority(self, key_id: str) -> Optional[_PUBLIC_KEY_TYPES]:
             None if the key_id is not found.
 
         Raises:
-            ValueError: When key_id is not valid (empty or None).
+            ValidationError: When key_id is not valid (empty or None).
```
maxlambrecht, line 64: For this kind of error, where an invalid argument was passed, I would throw an error named `IllegalArgumentError`, so it is immediately clear to the caller that some arguments were missing or empty.

src/pyspiffe/bundle/x509_bundle/x509_bundle.py
```diff
@@ -158,8 +158,8 @@ def load(
         if encoding == serialization.Encoding.DER:
             return cls.parse_raw(trust_domain, bundle_bytes)
 
-        raise ValueError(
-            'Encoding not supported: {}. Expected \'PEM\' or \'DER\'.'.format(encoding)
+        raise ValidationError(
```
maxlambrecht, line 161: This error is thrown when a not supported (aka illegal) encoding is passed, so it would raise a `IllegalArgumentError`

maxlambrecht (commented):
This looks very good @Andres-GC! I think the only change I'd suggest is renaming all `ValidationError`s by `IllegalArgumentError`s, making this way clear that the arguments were illegal/invalid, and that it is not some _internal_ validation error, that sounds obscure.

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
glaucimar: IllegalArgument seems java like.... ValueError seems more common in python codes for invalid arguments?

glaucimar (dismissed):

src/pyspiffe/bundle/jwt_bundle/jwt_bundle.py
maxlambrecht: Maybe `InvalidArgumentError`, `ArgumentError`, or `InvalidValueError`.  Something that doesn't have the word `Validation` in it.

Andres-GC: My vote is for `ArgumentError`.

maxlambrecht (dismissed):

glaucimar (dismissed):
Thanks!

maxlambrecht (approved):

glaucimar (approved):
Thanks!

Issue #67, opened 2021-03-31
Implement watch methods to get X.509 materials updates.


maxlambrecht:
Implement watch methods on Workload API client to get updates for X.509 materials asynchronously.

PR #84, opened 2021-04-19
Implement workload_api_client `watch_x509_context` method 

maxlambrecht:
This PR implements a _watch_ method on the Workload API client that keeps running fetching updates and allows to register two Callables to be executed either when there is a new update or when there is an error. 

It implements a retry functionality using an exponential backoff policy. This functionality can be turned off on the call of the _watch_ method. 

The _watch_ returns a CancelHandler to allow the caller to cancel the background process that fetches the updates. 

As part of this PR I also added a simple `close()` method on the Workload API client that closes the underlying channel.

I left a `#TODO` for implementing a customizable Backoff Retry Policy, in case we get to agree that is something worthy. 

Solved #67 


src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -81,6 +130,58 @@ def fetch_x509_svid(self) -> X509Svid:
...
+            retry_connect: Enable retries when the connection with the Workload API returns an error.
+                           Default: True
+
+        Returns:
+            CancelHandler: An object on which it can be called the method `cancel` to close the stream connection with
+                           the Workload API.
+        """
+
+        cancel_handler = CancelHandler()
+        retry_handler = _RetryHandler()
```
Andres-GC, line 172: _I might be missing something so sorry if it is the case here._
I see that this `retry_handler` is always initialized with and instance of `_RetryHandler()` no matter if `retry_connect` is true or false and then during the execution of this method both vars are passed as argument in several methods. I think that an alternative to avoid passing two arguments could be to initialize  `retry_handler` only when `retry_connect` is `True` (initialized to None in other case) and  then check whether _retry should occurs_ by confirming the presence of a retry handler with `if retry_handler`.

Andres-GC (commented):
Thanks Max.
I left a couple of minor comments.
Regarding the retry feature, I would follow a different approach. I think that is something we should have implemented globally and use it in all the calls to the workload API.

We can discuss it and I'm happy to take care of it if necessary.

test/workloadapi/test_default_workload_api_client_fetch_x509.py
```diff
@@ -562,3 +567,189 @@ def test_fetch_x509_bundles_corrupted_federated_bundle(mocker):
...
+    )
+
+    done = threading.Event()
+    response_holder = ResponseHolder()
+
+    _WORKLOAD_API_CLIENT.watch_x509_context(
+        lambda r: handle_x509_context_success(r, response_holder, done),
+        lambda e: handle_error(e, response_holder, done),
+        True,
+    )
```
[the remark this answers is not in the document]
maxlambrecht: Sure, it is a lot better using the parameter name.

test/workloadapi/test_default_workload_api_client_fetch_x509.py
```diff
@@ -562,3 +567,189 @@ def test_fetch_x509_bundles_corrupted_federated_bundle(mocker):
...
+    response_holder.success = response
+    event.set()
+
+
+def yield_grpc_error_and_then_correct_x509_svid_response():
+    grpc_error = grpc.RpcError()
+    grpc_error.code = lambda: grpc.StatusCode.DEADLINE_EXCEEDED
+    yield grpc_error
+
+    federated_bundles = dict()
```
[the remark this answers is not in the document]
maxlambrecht: Changed all the occurrences in that file.

src/pyspiffe/workloadapi/cancel_handler.py
```diff
@@ -0,0 +1,25 @@
+import threading
```
glaucimar, line 1: add module docstring.

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -34,6 +37,51 @@
...
+# GRPC Error Codes that the client will not retry on:
+#  - INVALID_ARGUMENT is not retried according to the SPIFFE spec because the request is invalid
+#  - CANCELLED is not retried because it occurs when the caller has canceled the operation.
+_NON_RETRYABLE_CODES = {grpc.StatusCode.CANCELLED, grpc.StatusCode.INVALID_ARGUMENT}
+
+
+class _RetryHandler:
+    def __init__(
+        self,
+        max_retries: int = 0,
```
glaucimar, line 49: should max_retries be set to another valid number? Like 3... that makes more sense the 0 to me.. Or the intent is not to have retries by default?

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -34,6 +37,51 @@
...
 )
 
+# GRPC Error Codes that the client will not retry on:
+#  - INVALID_ARGUMENT is not retried according to the SPIFFE spec because the request is invalid
+#  - CANCELLED is not retried because it occurs when the caller has canceled the operation.
+_NON_RETRYABLE_CODES = {grpc.StatusCode.CANCELLED, grpc.StatusCode.INVALID_ARGUMENT}
+
+
+class _RetryHandler:
+    def __init__(
```
glaucimar, line 47: I understand this is not required as this class is "internal" but I think it would be useful to have docstring for it.

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -34,6 +37,51 @@
...
+        self._max_retries = max_retries
+        self._base_backoff = base_backoff_in_seconds
+        self._backoff_factor = backoff_factor
+        self._max_delay_in_seconds = max_delay_in_seconds
+        self._retries_count = 0
+        self._lock = threading.RLock()
+
+    def do_retry(self, fn: Callable, params: List) -> bool:
+        with self._lock:
+            if self._max_retries and self._retries_count > self._max_retries:
```
glaucimar, line 63: As _retries_count starts with 0 and _max_retries is increased *after* this test, it should be
the test should be if self._max_retries and self._retries_count >= self._max_retries:
if retries_count = 0 and _max_retries = 1, do_retry could be called 3 times before falling in this if:
first time: retries_count = 0 and _max_retries = 1: self._retries_count > self._max_retries ? False
second time: retries_count = 1 and _max_retries = 1: self._retries_count > self._max_retries ? False
third time: retries_count = 2 and _max_retries = 1: self._retries_count > self._max_retries ? True

Only then, it would stop retries.... so, code would be executed twice instead of once as expected (_max_retries = 1).

src/pyspiffe/workloadapi/workload_api_client.py
```diff
@@ -34,6 +35,43 @@ def fetch_x509_svids(self) -> List[X509Svid]:
...
+           updates immediately after the Stream is ready and doesn't have to wait until the Workload API dispatches
+           the next update based on the SVIDs TTL.
+
+           In case of an error, if `retry_connect` is True and the error was not grpc.StatusCode.CANCELLED
+           or grpc.StatusCode.INVALID_ARGUMENT, it will attempt to establish a new connection
+           to the Workload API, using an exponential backoff policy to perform the retries, starting with a delay of 0.1 seconds,
+           incrementing it then to 0.2, 0.4, 0.8, 1.6 and so on (until the max backoff of 60 seconds). It retries indefinitely.
+
+        Args:
+            on_success: a Callable accepting a X509Context as argument and returning None, to be executed when a new update
```
glaucimar, line 61: on_sucess: A Callable..

the same for the other arguments below.

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -34,6 +37,51 @@
     WORKLOAD_API_HEADER_VALUE,
 )
 
+# GRPC Error Codes that the client will not retry on:
+#  - INVALID_ARGUMENT is not retried according to the SPIFFE spec because the request is invalid
+#  - CANCELLED is not retried because it occurs when the caller has canceled the operation.
+_NON_RETRYABLE_CODES = {grpc.StatusCode.CANCELLED, grpc.StatusCode.INVALID_ARGUMENT}
+
+
+class _RetryHandler:
```
glaucimar, line 46: I missed tests for this class.

glaucimar (commented):
Thanks for this, Max! Tons of work :). I have pointed out some items for your consideration.

src/pyspiffe/workloadapi/default_workload_api_client.py
maxlambrecht: The intent is to have an undefined number of retries by default, so if `max_retries=0` there is no maximum.

src/pyspiffe/workloadapi/default_workload_api_client.py
maxlambrecht: Good catch!

src/pyspiffe/workloadapi/default_workload_api_client.py
maxlambrecht: Added tests for that class and all the missing docs.

src/pyspiffe/workloadapi/default_workload_api_client.py
maxlambrecht: To make the intent clear I replaced this magic number "0" by a constant.

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -34,6 +37,69 @@
...
+
+
+class RetryHandler:
+    """Handler that performs retries using an exponential backoff policy."""
+
+    UNLIMITED_RETRIES = 0
+
+    def __init__(
+        self,
+        max_retries: int = UNLIMITED_RETRIES,
```
glaucimar, line 55: I was wondering if 0 should mean no retry at all and UNLIMITES_RETRIES should be -1.

glaucimar (commented):
Thanks Max! I have one final question :)

src/pyspiffe/workloadapi/default_workload_api_client.py
maxlambrecht: We don't need to codify _no retry_ in the RetryHandler since we would never create it in case we don't want retries.

maxlambrecht: ...and I read it as max-retries = 0 = False => no max at all, as in python _0_ evaluates as _False_

glaucimar: Makes sense.

glaucimar (approved):

maxlambrecht:
Some remaining points  will be addressed in https://github.com/HewlettPackard/py-spiffe/issues/91

PR #85, opened 2021-04-22
refactor code trying to avoid duplicate code

Andres-GC:
The PRs tries to avoid duplication of code by moving some methods to the `certificate_utils.py` module. The goals is to encapsulate all the logic required to read and write certs and keys to disk in only one place.

Let me know your thoughts please.

PR #88, opened 2021-04-28
Refactor 'save's as instance methods.

maxlambrecht:
Refactor the `save` methods in `X509Svid` and `X509Bundle` to be an instance methods, which makes a lot more sense than having a static method to be called through the class. So now the call to save an `X509Svid` would look like:

`my_svid.save('/tmp/my_cert.pem', /tmp/my_key.pem', encoding.PEM)`


Andres-GC (commented):
These changes have conflicts with changes on this previous pending PR: https://github.com/HewlettPackard/py-spiffe/pull/85

Let's merged the other PR before merging this one.

Andres-GC (approved):
🍦

PR #92, opened 2021-05-12
Add missing validation in TrustDomain

maxlambrecht:
The parse TrustDomain method is currently allowing parsing strings that include userinfo (e.g. `user:pass@domain.test`). Added a validation to prevent that.


Andres-GC (approved):
💯

Issue #69, opened 2021-03-31
Define and implement a X.509 Source interface.


maxlambrecht:
Implement a source of X.509 materials that keeps them in memory up to date with the most recent updates pushed by the Workload API. This is a convenient class to handle X.509 svid, private key and bundles in memory.

PR #93, opened 2021-05-12
Implement X.509 Source

maxlambrecht:
This is the implementation of an X.509 Source that is kept updated by an underlying Workload API client connection that retrieves the X.509 context updates.

As part of this implementation some minor changes were made on the Workload API client with respect to the report of the error to the clients on the `watch_x509_context` method. Also some validations were added on the constructor of the X509Context.

Solves #69 

### DIFF
--- a/src/pyspiffe/workloadapi/default_workload_api_client.py
+++ b/src/pyspiffe/workloadapi/default_workload_api_client.py
@@ -1,6 +1,7 @@
 """
 This module provides a Workload API client.
 """
+import logging
 import threading
 import time
 from typing import Optional, List, Mapping, Iterator, Callable
@@ -177,7 +178,7 @@ class DefaultWorkloadApiClient(WorkloadApiClient):
                         is fetched from the Workload API.
 
             on_error: A Callable accepting an Exception as argument and returning None, to be executed when there is
-                      an error on the connection with the Workload API.
+                      an error on the connection with the Workload API after what there is no retries.
 
             retry_connect: Enable retries when the connection with the Workload API returns an error.
                            Default: True.
@@ -474,11 +475,16 @@ class DefaultWorkloadApiClient(WorkloadApiClient):
         on_error: Callable[[Exception], None],
     ):
         grpc_error_code = grpc_error.code()
-        error = FetchX509SvidError(str(grpc_error_code))
-        on_error(error)
 
         if retry_handler and grpc_error_code not in _NON_RETRYABLE_CODES:
+            logging.error(
+                'Error connecting to the Workload API: {}'.format(str(grpc_error_code))
+            )
             retry_handler.do_retry(
                 self._call_watch_x509_context,
                 [cancel_handler, retry_handler, on_success, on_error],
             )
+        else:
+            # don't retry, instead report error to user on the on_error callback
+            error = FetchX509SvidError(str(grpc_error_code))
+            on_error(error)
--- a/src/pyspiffe/workloadapi/default_x509_source.py
+++ b/src/pyspiffe/workloadapi/default_x509_source.py
@@ -1,0 +1,129 @@
+import logging
+import threading
+from contextlib import suppress
+from typing import Optional, Callable, List
+
+from pyspiffe.bundle.x509_bundle.x509_bundle import X509Bundle
+from pyspiffe.spiffe_id.trust_domain import TrustDomain
+from pyspiffe.svid.x509_svid import X509Svid
+from pyspiffe.workloadapi.default_workload_api_client import DefaultWorkloadApiClient
+from pyspiffe.workloadapi.exceptions import X509SourceError
+from pyspiffe.workloadapi.workload_api_client import WorkloadApiClient
+from pyspiffe.workloadapi.x509_context import X509Context
+from pyspiffe.workloadapi.x509_source import X509Source
+
+
+class DefaultX509Source(X509Source):
+    """Source of X509-SVIDs and X.509 bundles maintained via the Workload API."""
+
+    def __init__(
+        self,
+        workload_api_client: WorkloadApiClient = None,
+        spiffe_socket: str = None,
+        timeout_in_seconds: float = None,
+        picker: Callable[[List[X509Svid]], X509Svid] = None,
+    ) -> None:
+        """Creates a new X509Source.
+
+           It blocks until the initial update has been received from the Workload API or until timeout_in_seconds is reached.
+
+           In case the underlying Workload API connection returns an unretryable error, the source will be closed and
+           no methods on the source will be available.
+
+
+        Args:
+            workload_api_client: A WorkloadApiClient that will be used to fetch the X.509 materials from the Workload API.
+                                 In case it's not provided, a default client will be created.
+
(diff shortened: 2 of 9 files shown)


src/pyspiffe/workloadapi/exceptions.py
```diff
@@ -75,3 +75,17 @@ def __init__(self, additional_information: str = 'none') -> None:
             additional_information: Additional information about the error.
         """
         super().__init__(self._MESSAGE.format(additional_information))
+
+
+class X509SourceError(WorkloadApiError):
+    """Error raised when there is an error in the X.509 Source."""
+
+    _MESSAGE = 'X.509 Source error: {}.'
```
Andres-GC, line 83: nit: final dot is set automatically so we removed it from all the custom exceptions.

src/pyspiffe/workloadapi/default_x509_source.py
```diff
@@ -0,0 +1,129 @@
...
+                raise X509SourceError('Cannot get X.509 Bundle: source is closed')
+        return self._x509_bundle_set.get_x509_bundle_for_trust_domain(trust_domain)
+
+    def close(self) -> None:
+        """Closes this X509Source closing the underlying connection with the Workload API. Once the source is closed,
+        no methods can be called on it.
+        """
+        with self._lock:
+            # the cancel method throws a grpc exception, that can be discarded
+            with suppress(Exception):
```
glaucimar, line 104: can any other exception be thrown that we should care about? In any case, we should  add a log entry for the raised exception.

src/pyspiffe/workloadapi/x509_source.py
```diff
@@ -0,0 +1,27 @@
+from abc import ABC, abstractmethod
```
glaucimar, line 1: add module docstring

glaucimar (commented):
Great work, @maxlambrecht

src/pyspiffe/workloadapi/default_x509_source.py
```diff
@@ -0,0 +1,129 @@
...
+        self, trust_domain: TrustDomain
+    ) -> Optional[X509Bundle]:
+        """Returns the X.509 bundle for the given trust domain."""
+        with self._lock:
+            if self._closed:
+                raise X509SourceError('Cannot get X.509 Bundle: source is closed')
+        return self._x509_bundle_set.get_x509_bundle_for_trust_domain(trust_domain)
+
+    def close(self) -> None:
+        """Closes this X509Source closing the underlying connection with the Workload API. Once the source is closed,
```
glaucimar, line 99: let's make crystal clear that the user has the responsibility of calling this method to close the underlying connection to WL api. I think we must make this point in order to warn user that tit is his responsibility of doing so and "saving" resources.

src/pyspiffe/workloadapi/default_x509_source.py
```diff
@@ -0,0 +1,129 @@
...
+            if self._picker:
+                self._x509_svid = self._picker(x509_context.x509_svids())
+            else:
+                self._x509_svid = x509_context.default_svid()
+
+            self._x509_bundle_set = x509_context.x509_bundle_set()
+            self._initialized.set()
+        except Exception as err:
+            logging.error(
+                'X.509 Source: error setting X.509 context: {}.'.format(str(err))
```
glaucimar, line 121: Should this be error _getting_ X.509 context? or material? Seems like the error comes most probably from the calls to context then from the assignment to the variables.. although, I do get the point that this methoding is getting updated information... maybe something like: error updating X.509 material?

src/pyspiffe/workloadapi/default_x509_source.py
```diff
@@ -0,0 +1,129 @@
...
+            raise X509SourceError(
+                'Could not initialize X.509 Source: reached timeout waiting for the first update'
+            )
+
+    def get_x509_svid(self) -> X509Svid:
+        """Returns an X509-SVID from the source."""
+        with self._lock:
+            if self._closed:
+                raise X509SourceError('Cannot get X.509 SVID: source is closed')
+        return self._x509_svid
```
glaucimar, line 87: access to x509_svid should be thread safe too, this should be inside the with block.

src/pyspiffe/workloadapi/default_x509_source.py
```diff
@@ -0,0 +1,129 @@
...
+            # the cancel method throws a grpc exception, that can be discarded
+            with suppress(Exception):
+                self._client_cancel_handler.cancel()
+            # prevents blocking on the constructor
+            self._initialized.set()
+            self._closed = True
+
+    def _set_context(self, x509_context: X509Context) -> None:
+        try:
+            if self._picker:
```
glaucimar, line 112: _x509_svid and _x509_bundle_set access should be thread safe, this access should be protected too.

src/pyspiffe/workloadapi/default_x509_source.py
```diff
@@ -0,0 +1,129 @@
...
+        return self._x509_svid
+
+    def get_bundle_for_trust_domain(
+        self, trust_domain: TrustDomain
+    ) -> Optional[X509Bundle]:
+        """Returns the X.509 bundle for the given trust domain."""
+        with self._lock:
+            if self._closed:
+                raise X509SourceError('Cannot get X.509 Bundle: source is closed')
+        return self._x509_bundle_set.get_x509_bundle_for_trust_domain(trust_domain)
```
glaucimar, line 96: access to _x509_bundle_set should be thread safe, this should be protected too.

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -474,11 +475,16 @@ def _handle_grpc_error(
         on_error: Callable[[Exception], None],
     ):
         grpc_error_code = grpc_error.code()
-        error = FetchX509SvidError(str(grpc_error_code))
-        on_error(error)
 
         if retry_handler and grpc_error_code not in _NON_RETRYABLE_CODES:
+            logging.error(
+                'Error connecting to the Workload API: {}'.format(str(grpc_error_code))
+            )
```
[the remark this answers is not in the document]
maxlambrecht: I don't know what options you are referring to.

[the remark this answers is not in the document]
Andres-GC: Here we are using the default logging module directly. I think we should create our own logger and configure it.
By default, the `logging` module logs the messages with a severity level of `WARNING` or above. This does not impact here because logs are for error and exception but if we want to add logs for debug purposes we need to set the _LogLevel_ accordingly. 
The default format for those messages is: `LEVEL:logger:message`. I think we can add more contexts to the log statements like time, module or line number. Then if necessary we can also write logs in json format. There are several options to consider.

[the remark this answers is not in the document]
maxlambrecht: Can you create a ticket to address the logging properties ?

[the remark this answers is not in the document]
Andres-GC: Sure, this is the issue: https://github.com/HewlettPackard/py-spiffe/issues/95

Andres-GC (approved):

PR #94, opened 2021-05-19
Adds watch_jwt_bundle feature

Andres-GC:


src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -295,16 +295,50 @@ def fetch_jwt_bundles(self) -> JwtBundleSet:
...
+    def watch_jwt_bundles(
+        self,
+        on_success: Callable[[JwtBundleSet], None],
+        on_error: Callable[[Exception], None],
+        retry_connect: bool = True,
+    ) -> CancelHandler:
+        """Watches for changes to the JWT bundles.
+
+        Args:
+            on_success: A Callable accepting a X509Context as argument and returning None, to be executed when a new
```
glaucimar, line 314: it should be "accepting a JwtBundleSet" instead of X509Context

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -295,16 +295,50 @@ def fetch_jwt_bundles(self) -> JwtBundleSet:
...
+                        update is fetched from the Workload API.
+
+            on_error: A Callable accepting an Exception as argument and returning None, to be executed when there is
+                      an error on the connection with the Workload API.
+
+            retry_connect: Enable retries when the connection with the Workload API returns an error. Default: True.
+
+        Returns:
+            CancelHandler: An object on which it can be called the method `cancel` to close the stream connection with
+                           the Workload API.
```
glaucimar, line 324: Suggestion: call out that the client is responsible to call the cancel method to stop the connection whenever this is not required to avoid using resources when not needed.

src/pyspiffe/workloadapi/workload_api_client.py
```diff
@@ -111,6 +111,29 @@ def fetch_jwt_bundles(self) -> JwtBundleSet:
             JwtBundleSet: Set of JwtBundle objects.
         """
 
+    @abstractmethod
+    def watch_jwt_bundles(
+        self,
+        on_success: Callable[[X509Context], None],
```
glaucimar, line 117: it should be Callable[[JwtBundleSet]]

src/pyspiffe/workloadapi/workload_api_client.py
```diff
@@ -111,6 +111,29 @@ def fetch_jwt_bundles(self) -> JwtBundleSet:
...
+        on_error: Callable[[Exception], None],
+        retry_connect: bool = True,
+    ) -> CancelHandler:
+        """Watches for changes to the JWT bundles.
+
+        Args:
+            on_success: A Callable accepting a X509Context as argument and returning None, to be executed when a new
+                        update is fetched from the Workload API.
+
+            on_error: A Callable accepting an Exception as argument and returning None, to be executed when there is
```
glaucimar, line 127: JwtBundleSet instead of X509Context

src/pyspiffe/workloadapi/default_workload_api_client.py
Andres-GC: Thanks Glaucimar.
I've added the text Max used for the other watcher method so it is consistent.

glaucimar (approved):

maxlambrecht (approved):
🏅

Issue #95, opened 2021-05-20
Configure Logging 


Andres-GC:
Define and create a logging configuration for the library.
Things to considerer:
- Log Level should be configurable
- Format of the log statements
- Should the library generates logs in json format?

PR #98, opened 2021-06-02
Adds logging configuration

Andres-GC:
The configuration creates a new logger for the library with a `NullHandler` to avoid any message to be logged.
This allows the library to be integrated with the library's client logging configuration.

If required, a client of the library can set the log level by:
```
py_spiffe_logger = logging.getLogger('pyspiffe')
py_spiffe_logger.setLevel(logging.DEBUG)
``` 
On the other hand, inside the library we create a logger per module using:
`_logger = logging.getLogger(__name__)`

This allows us to get a logger with a name that follows the library architecture. For example, when getting a logger at `default_workload_api_client.py` we end up having a log record from a logger called `pyspiffe.workloadapi.default_workload_api_client`.

Solves https://github.com/HewlettPackard/py-spiffe/issues/95

### DIFF
--- a/src/pyspiffe/__init__.py
+++ b/src/pyspiffe/__init__.py
@@ -8,3 +8,7 @@ TODO: add usage example here
    >>>
 
 """
+import logging
+from logging import NullHandler
+
+logging.getLogger(__name__).addHandler(NullHandler())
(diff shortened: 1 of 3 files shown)


maxlambrecht (approved):
💯

glaucimar (commented):
Should this behavior be documented somewhere?

Andres-GC:
> Should this behavior be documented somewhere?

This is the common behavior for Python libraries so I'm not sure if we want something specific. What I think we can do is to included it as part of the examples.

Issue #70, opened 2021-03-31
Define and implement a JWT Source interface.


maxlambrecht:
Implement a source of JWT materials that keeps them in memory up to date with the most recent updates pushed by the Workload API. This is a convenient class to handle JWT bundles in memory.

glaucimar:
done

PR #99, opened 2021-06-03
Implements JWT Source

glaucimar:
This addresses issue #70 .

Creates JWT Source and Default JWT Source implementation.


src/pyspiffe/workloadapi/jwt_source.py
```diff
@@ -0,0 +1,36 @@
...
+class JwtSource(ABC):
+    """Source of JWT-SVIDs and JWT bundles maintained via the Workload API."""
+
+    @abstractmethod
+    def get_jwt_svid(self) -> JwtSvid:
+        """Returns an JWT-SVID from the source."""
+        pass
+
+    @abstractmethod
+    def get_bundle_for_trust_domain(
```
Andres-GC, line 22: I think we should review the name of this method and get an agreement on how to name it.
The `JwtBundleSet` use `get()` instead of `get_bundle_for_trust_domain()` but then all the methods from the x509 side use the latter.
I remember having a conversation to define that `JwtBundleSet` should have a `get()` method because is pretty much a dictionary keyed by Trust Domain. I still prefer the `get()` name but I would like to be consistent on this name so if you think that the name here should remains as `get_bundle_for_trust_domain()` let's rename the method on `JwtBundleSet`. What do you think?

src/pyspiffe/workloadapi/exceptions.py
```diff
@@ -103,3 +103,16 @@ def __init__(self, additional_information: str) -> None:
...
+
+
+class JwtSourceError(WorkloadApiError):
+    """Error raised when there is an error in the JWT Source."""
+
+    _MESSAGE = 'JWT Source error: {}'
+
+    def __init__(self, additional_information: str) -> None:
+        """Creates an instance of JWTSourceError.
+        Args:
```
Andres-GC, line 115: nit: missing a blank line here.

src/pyspiffe/workloadapi/default_jwt_source.py
```diff
@@ -0,0 +1,149 @@
...
+from pyspiffe.spiffe_id.trust_domain import TrustDomain
+from pyspiffe.svid.jwt_svid import JwtSvid
+from pyspiffe.workloadapi.default_workload_api_client import DefaultWorkloadApiClient
+from pyspiffe.workloadapi.workload_api_client import WorkloadApiClient
+from pyspiffe.workloadapi.jwt_source import JwtSource
+from pyspiffe.workloadapi.exceptions import JwtSourceError
+
+
+class DefaultJwtSource(JwtSource):
+    """Default implementation of JwtSource. This class may be used by clients to get updated JWT-SVIDs and JWT bundles without the need to contact WorkloadAPI.
```
maxlambrecht, line 20: This is true for the bundles, but for the JWT-SVID it will connect to the Workload API each time as the client may need a token with different audience.

src/pyspiffe/workloadapi/default_jwt_source.py
```diff
@@ -0,0 +1,149 @@
...
+    DefaultJwtSource treates updates from WorkloadAPI and keeps JWT-SVID and JWT bundles updated.
```
maxlambrecht, line 21: `Workload API`

src/pyspiffe/workloadapi/default_jwt_source.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+
+    def __init__(
+        self,
+        audiences: Set[str],
```
maxlambrecht, line 26: The parameters `audience` and `subject` will not be required here, they should be parameters of a `fetch_jwt_svid` method.

src/pyspiffe/workloadapi/default_jwt_source.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+        self._initialized.wait(timeout_in_seconds)
+
+        if not self._initialized.is_set():
+            self._client_cancel_handler.cancel()
+            raise JwtSourceError(
+                'Could not initialize JWT Source: reached timeout waiting for the first update'
+            )
+
+    def get_jwt_svid(self) -> JwtSvid:
```
maxlambrecht, line 77: This method should take `audience` and `subject` as parameters and fetch the token, thus `fetch_jwt_svid` would be more appropriate.

maxlambrecht (changes requested):

src/pyspiffe/workloadapi/jwt_source.py
glaucimar: I  think you are correct. I will rename it.. I guess I forgot that conversations :S...:)

Thanks for catching this!

src/pyspiffe/workloadapi/default_jwt_source.py
glaucimar: hmmmmm that is a good catch Max, thanks! I will make the change and make audience a method parameter.

src/pyspiffe/workloadapi/jwt_source.py
maxlambrecht: Now that I see the code again, it feels a little confusing the name `get`, as the source not only provides bundles but also jwt-svids, it seems to me that something like `get_bundle` or `get_jwt_bundle` would clarify better the purpose of the method. Sorry for bringing this up now.

src/pyspiffe/workloadapi/default_jwt_source.py
```diff
@@ -0,0 +1,150 @@
...
+        try:
+            jwt_svid = self._workload_api_client.fetch_jwt_svid(audiences, subject)
+            return jwt_svid
+        except Exception as err:
+            logging.error('JWT Source: error getting JWT SVID: {}.'.format(str(err)))
+            logging.error('JWT Source: closing due to invalid state.')
+            self.close()
+            raise JwtSourceError(
+                'JWT Source: error getting JWT SVID: {}'.format(str(err))
+            )
```
Andres-GC, line 101: I'm wondering if any error should cause the source to be closed? 
`fetch_jwt_svid()` is a one-shot call that can return a validation error (like SPIFFE ID malformed or invalid) and in that case I think it might be reasonable to return an error without closing the source. Does it make sense?

If it does, we have another issue that is how to identify that kind of errors because `fetch_jwt_svid()` will raise `FetchJwtSvidError` in several cases. We need to think about this if necessary.

src/pyspiffe/workloadapi/default_jwt_source.py
```diff
@@ -0,0 +1,150 @@
...
+            self._jwt_bundle_set = jwt_bundle_set
+            self._initialized.set()
+
+    def _on_error(self, error: Exception) -> None:
+        self._log_error(error)
+        self.close()
+
+    def _log_error(self, err: Exception) -> None:
+        logging.error('JWT Source: Workload API client error: {}'.format(str(err)))
+        logging.error('JWT Source: closing.')
```
Andres-GC, line 150: Based on some changes introduced after you started this PR, you'll need to update the way you use the logger.
You need to define the logger for the module and then use it to log your messages.

```
_logger = logging.getLogger(__name__)
...
_logger.error('JWT Source: Workload API client error: {}'.format(str(err)))
```

src/pyspiffe/workloadapi/default_jwt_source.py
glaucimar: Hi Andres, thanks! I agree... I have removed the close and removed the catch Exception :).

maxlambrecht (approved):
🥇

Andres-GC (approved):
Thanks Glaucimar!

PR #100, opened 2021-06-09
Unify error handling on default workload API

Andres-GC:
This PRs unified the error handling on the `default_workload_api_client.py`.
This includes checking if the error response is a `grpc.Call` which contains an error Code and Details. If it is the case, the details are included in the exception thrown to the user of the library.


maxlambrecht (approved):
This looks really nice! 🎉

PR #103, opened 2021-07-30
Fix validate path method.

maxlambrecht:
This PR does some amendments on the `SpiffeId` parsing and validation of paths. These same changes have been introduced to `go-spiffe` and `java-spiffe` libraries. I also changed the method name for creating spiffe_ids from segments to make it more clear. 


glaucimar (approved):

Andres-GC (approved):
:100: 
Thanks Max.

PR #109, opened 2024-03-05
Add a fix to the error raised when using empty spiffeID in fetch_jwt_…

telliere:
…svid

Modifying this part :

```python
    @handle_error(error_cls=FetchJwtSvidError)
    def fetch_jwt_svid(
        self, audiences: List[str], subject: Optional[SpiffeId] = None
    ) -> JwtSvid:
        if not audiences:
            raise ArgumentError('Parameter audiences cannot be empty')
       
        response = self._spiffe_workload_api_stub.FetchJWTSVID(
            request=workload_pb2.JWTSVIDRequest(
                audience=audiences,
                spiffe_id=str(subject),
            )
        )

        if len(response.svids) == 0:
            raise FetchJwtSvidError('JWT SVID response is empty')

        svid = response.svids[0].svid
        return JwtSvid.parse_insecure(svid, audiences)
```

Bug :

When using fetch_jwt_svid with no spiffeID specified it raises a `FetchJwtSvidError` while it shouldn't.

Cause :

```
spiffe_id=str(subject),
```

parsing a `None` object as a str creates `'None'`, causing the next function (`JWTSVIDRequest`) to consider having a specified spiffeID when None has been specified, handled by the `@handle_error` as a `FetchJwtSvidError`.

Fix :

Handling the NoneType before, by casting `subject` to a `str` only if it's not None.

```
        if subject != None:
            subject = str(subject)
```

### DIFF
--- a/src/pyspiffe/workloadapi/default_workload_api_client.py
+++ b/src/pyspiffe/workloadapi/default_workload_api_client.py
@@ -279,10 +279,13 @@ class DefaultWorkloadApiClient(WorkloadApiClient):
         if not audiences:
             raise ArgumentError('Parameter audiences cannot be empty')
 
+        if subject != None:
+            subject = str(subject)
+
         response = self._spiffe_workload_api_stub.FetchJWTSVID(
             request=workload_pb2.JWTSVIDRequest(
                 audience=audiences,
-                spiffe_id=str(subject),
+                spiffe_id=subject,
             )
         )
 
--- a/src/pyspiffe/workloadapi/default_workload_api_client.py
+++ b/src/pyspiffe/workloadapi/default_workload_api_client.py
@@ -283,7 +283,7 @@ class DefaultWorkloadApiClient(WorkloadApiClient):
         response = self._spiffe_workload_api_stub.FetchJWTSVID(
             request=workload_pb2.JWTSVIDRequest(
                 audience=audiences,
-                spiffe_id=subject,
+                spiffe_id=subject_str,
             )
         )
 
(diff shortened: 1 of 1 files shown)


maxlambrecht:
Thanks for this PR, @telliere!

Would you rebase your branch?

telliere:
Done,

Actually I have to say that the other PR opened would help me a lot, but I didn't have a look at the code

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -279,10 +279,13 @@ def fetch_jwt_svid(
         if not audiences:
             raise ArgumentError('Parameter audiences cannot be empty')
 
+        if subject is not None:
+            subject = str(subject)
+
```
maxlambrecht, line 284: We are getting this error: `src/pyspiffe/workloadapi/default_workload_api_client.py:283: error: Incompatible types in assignment (expression has type "str", variable has type "Optional[SpiffeId]")  [assignment]`

```suggestion
        subject = str(subject) if subject is not None else ''
```

telliere:
Was of the subject, sorry about that. Thanks for the review / The suggestions

src/pyspiffe/workloadapi/default_workload_api_client.py
```diff
@@ -279,10 +279,11 @@ def fetch_jwt_svid(
         if not audiences:
             raise ArgumentError('Parameter audiences cannot be empty')
 
+        subject_str = str(subject) if subject is not None else ''
         response = self._spiffe_workload_api_stub.FetchJWTSVID(
             request=workload_pb2.JWTSVIDRequest(
                 audience=audiences,
-                spiffe_id=str(subject),
+                spiffe_id=subject,
```
maxlambrecht, line 286:
```suggestion
                spiffe_id=subject_str,
```

PR #412, opened 2026-03-18
fix: reject X.509-SVID leaf SPIFFE IDs with root path

maxlambrecht:
## Summary

Enforces X.509-SVID spec: `Updates X509Svid.parse` and `X509Svid.parse_raw` so that leaf certificates are rejected when their URI SAN is a SPIFFE ID with an empty/root path (e.g. spiffe://example.org).
Leaf-only restriction: The new check runs only on the extracted SPIFFE ID for the leaf certificate; intermediate/CA certificates are unaffected.
Error semantics: When a leaf SPIFFE ID has no path, `InvalidLeafCertificateError` is raised with the message: Invalid leaf certificate: Leaf certificate SPIFFE ID must have a non-root path.

### DIFF
--- a/spiffe/CHANGELOG.md
+++ b/spiffe/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## [Unreleased]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


amartinezfayo (approved):
Looks great @maxlambrecht!
I left some minor suggestions.

amartinezfayo (approved):
🎉

PR #414, opened 2026-03-19
fix(spiffe_id): accept mixed-case SPIFFE scheme and trust domain

maxlambrecht:
## What
Make SPIFFE ID URI scheme matching case-insensitive.
Accept mixed-case trust domains, normalize them to lowercase, then validate.
Preserve the path exactly as provided (case-sensitive) and keep existing invalid-path and invalid-trust-domain rejections.

## Why
Align parsing behavior with the SPIFFE ID spec and allow valid mixed-case inputs while keeping a canonical internal trust-domain representation.

## How tested

Updated/added unit tests for scheme/trust-domain mixed-case acceptance, lowercase canonicalization, path case preservation, and continued rejection of invalid trust-domain characters.


maxlambrecht:
Thanks for the review, @amartinezfayo 

Great catches.

I addressed all three points:

- allowed underscores in trust-domain labels per SPIFFE spec guidance
- added assert a != "EXAMPLE.ORG" in the trust-domain canonicalization test
- added a path case-sensitivity inequality check (/Service vs /service) for SpiffeId

PR #420, opened 2026-04-13
fix: update uv.lock and configure dependabot to use uv

arpad-csepi:
**Pull Request check list**

- [X] Commit conforms to CONTRIBUTING.md?
- [X] Proper tests/regressions included?
- [X] Documentation updated?

**Affected functionality**
Dependabot PRs about package upgrades are not upgrade the `uv.lock` file to keep in sync with `pyproject.toml`.
This causes a package to be stuck on the older version and the newer is not tested in the CI.

**Description of change**
- Replaced pip with uv value in `package-ecosystem` inside of the dependabot config, so dependabot will keep `uv.lock` file in sync with the `pyproject.toml` and the newer version of the package will be used in testing as well on CI.
- Removed the `uv.lock` and regenerated with `uv lock` command. Some of the package versioned are narrowed down (but can be dropped from commits if does not needed).

Source: https://docs.astral.sh/uv/guides/integration/dependabot/

If dependabot does not working correctly then might worth to migrate to renovatebot which has better uv support.

**Which issue this PR fixes**


maxlambrecht (approved):
LGTM, approved. 
Let’s verify in the next Dependabot cycle that root uv.lock updates consistently with the new uv ecosystem config.

PR #426, opened 2026-05-07
chore(deps): remove cryptography upper bound

maxlambrecht:
**Affected functionality**

Runtime dependency constraints for the `spiffe` package.

**Description of change**

Removes the upper bound on the `cryptography` runtime dependency:

- changes `cryptography>=46,<48` to `cryptography>=46`
- refreshes `uv.lock` metadata

As a library, `spiffe` should avoid blocking downstream applications from upgrading `cryptography` unless there is a known incompatibility. CI and Dependabot can catch future breakage, and a targeted upper bound can be reintroduced if needed.

### DIFF
--- a/spiffe/pyproject.toml
+++ b/spiffe/pyproject.toml
@@ -8,7 +8,7 @@ requires-python = ">=3.10"
 authors = [{ name = "Max Lambrecht", email = "maxlambrecht@gmail.com" }]
 dependencies = [
   "grpcio>=1.78.0,<2",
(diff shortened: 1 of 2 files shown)


maxlambrecht:
Hey @sorindumitru, thanks for the suggestion on the Dependabot PR #425. This follows up by removing the upper bound on `cryptography` rather than continuing to bump it for each major release.

sorindumitru (commented):
Thanks for the quick response here, really appreciate it. This looks good to me

PR #443, opened 2026-07-09
Disable 'iat' field verification

sorindumitru:
**Pull Request check list**

- [ ] Commit conforms to CONTRIBUTING.md?
- [ ] Proper tests/regressions included?
- [ ] Documentation updated?

**Affected functionality**
JWT-SVID verification

**Description of change**
The pyjwt library verified the 'iat' field [by default](https://github.com/jpadilla/pyjwt/issues/939) even though it's somewhat wrong.

Since this can lead to "token not yet valid", could we disable that extra verification?

### DIFF
--- a/spiffe/src/spiffe/svid/jwt_svid.py
+++ b/spiffe/src/spiffe/svid/jwt_svid.py
@@ -176,6 +176,7 @@ class JwtSvid(object):
                     'verify_signature': True,
                     'verify_aud': True,
                     'verify_exp': True,
+                    'verify_iat': False,
                 },
             )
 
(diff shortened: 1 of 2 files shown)


maxlambrecht:
Thanks, @sorindumitru. This behavior change makes sense to me. `iat` is not part of the JWT-SVID validation requirements, and relying on PyJWT’s default `iat` validation can reject otherwise valid JWT-SVIDs under normal clock skew.

Could you add a focused regression test for this? Something in `test_jwt_svid.py` that signs a valid JWT-SVID with an `iat` a few minutes in the future and asserts that `JwtSvid.parse_and_validate()` succeeds. That will lock in the intended behavior while preserving signature, audience, and expiration validation.

sorindumitru:
@maxlambrecht, I've added a test for this, tested it with `verify_iat: True` and it throws with:
```
E           spiffe.svid.errors.InvalidTokenError: The token is not yet valid (iat)
```
so it seems to be doing the right thing.

maxlambrecht (approved):
LGTM! 

Thanks, @sorindumitru !

